PR #11, opened 2019-09-26
feat: add a babel-preserve-modules config preset

abutterworth:


config/babel-preserve-modules.config.js
```diff
@@ -0,0 +1,34 @@
...
+          {
+            messagesDir: './temp/babel-plugin-react-intl',
+            moduleSourceName: '@edx/frontend-i18n',
+          },
+        ],
+      ],
+    },
+    test: {
+      presets: [
+        '@babel/preset-env',
```
davidjoy: Don't you need the react preset here too?  Or does this get overlaid on top of the normal presets?

abutterworth: it gets overlaid

abutterworth: https://babeljs.io/docs/en/options#env

PR #13, opened 2019-10-11
Process favicon.ico and a PROJECT_ROOT fix

davidjoy:


lib/paths.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 
 module.exports = {
-  PROJECT_ROOT: process.env.PROJECT_ROOT,
+  PROJECT_ROOT: process.env.PROJECT_ROOT ? process.env.PROJECT_ROOT : process.cwd(),
```
davidjoy: Had this sitting in my checkout as a small fix @abutterworth and I talked about - threw it in to save tedium.

davidjoy: process.env.PROJECT_ROOT can be undefined if you use any of this code outside of calling it through `fedx-scripts` - this is common if you're trying to extend a webpack config, for instance.

PR #42, opened 2020-01-30
fix: updating to use @edx/eslint-config

davidjoy:
BREAKING CHANGE: Marking this as a breaking change as it’s likely to break a number of linting rules until we find the right overrides.

### DIFF
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -1,5 +1,5 @@
 module.exports = {
-  "extends": "eslint-config-edx",
+  "extends": "@edx/eslint-config",
   "parser": "babel-eslint",
   "rules": {
     "import/no-extraneous-dependencies": [
@@ -17,16 +17,6 @@ module.exports = {
       "components": [ "Link" ],
       "specialLink": [ "to" ]
     }],
-    "jsx-a11y/label-has-for": [ 2, {
-      "components": [ "label" ],
-      "required": {
-        "some": [ "nesting", "id" ]
-      },
-      "allowChildren": false
-    }]
-  },
-  "env": {
-    "jest": true
   },
   "globals": {
     "newrelic": false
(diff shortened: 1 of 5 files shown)


config/.eslintrc.js
```diff
@@ -17,16 +17,6 @@ module.exports = {
       "components": [ "Link" ],
       "specialLink": [ "to" ]
     }],
-    "jsx-a11y/label-has-for": [ 2, {
```
davidjoy: This is a deprecated warning.

package.json
```diff
@@ -45,8 +46,12 @@
     "cssnano": "4.1.10",
     "dotenv": "8.1.0",
     "dotenv-webpack": "1.7.0",
-    "eslint": "6.4.0",
-    "eslint-config-edx": "4.0.4",
+    "eslint": "6.8.0",
```
davidjoy: eslint-config-airbnb defines it's dependencies as peer dependencies - our @edx/eslint-config now does the same.  That means frontend-build - as the user of our config - is responsible for installing the peer deps.  An MFE that used @edx/eslint-config directly would have to do the same thing, which is the behavior of eslint-config-airbnb too.

davidjoy: It's a bit odd, as eslint-config-airbnb expects them to be installed as dev dependencies, which would normally make sense.  In the case of frontend-build, it _feels_ as if they should be installed as full dependencies since this is a build-time tool.  We may eventually iterate on this...

config/.eslintrc.js
abutterworth: what do you mean? It doesn't apply anymore?

package.json
abutterworth: I think you're right to make them full dependencies. Otherwise they wouldn't be installed when frontend-build is added as a dev dependency of an MFE.

config/.eslintrc.js
davidjoy: To my knowledge it doesn’t get checked anymore and has been replaced by some other rules. At least that’s what I thought I read!

PR #47, opened 2020-03-21
Update cli passthrough and add smoke test

abutterworth:
Updating the way CLI passthrough happens that will enabled smoother development and makes automated testing easier to do. frontend-build no longer relies upon npm to have properly created or moved aliases of scripts in other packages to the `.bin/` directory. Instead it relies upon the fact that these scripts must all live inside their packages and requires them directly there. For example: `require('webpack/bin/webpack');` is the same script triggered by `npm run webpack`


.travis.yml
```diff
@@ -4,6 +4,7 @@ install:
 - npm install
 script:
 - npm run lint
+- npm run test
```
abutterworth: Our new smoke test (smoke-test.sh)

README.rst
```diff
@@ -88,28 +88,33 @@ line if you prefer a different location. Example package.json::
...
      "scripts": {
-        "build": "fedx-scripts webpack --config=./config/webpack.config.js",
-        "start:stage": "fedx-scripts webpack-dev-server --config=webpack.dev-stage.config.js",
+        "build": "fedx-scripts webpack --config ./config/webpack.config.js",
+        "start:stage": "fedx-scripts webpack-dev-server --config webpack.dev-stage.config.js",
         ...
      }
   }
 
+Note, specifying a custom config location for babel may cause issues with other
```
abutterworth: I noticed this existing limitation while working through stuff. It's addressable by adding some custom config parameters, but I think we can wait until someone actually needs to do this.

README.rst
```diff
@@ -88,28 +88,33 @@ line if you prefer a different location. Example package.json::
...
-dependencies in the right location, but it's not fool proof. If you run into errors that
-say something like "webpack: command not found" you have two options. 
-
-1. Delete the node_modules directories from the host project and this one and re-run 
-``npm i -D ./frontend-build`` from the host project.
-
-2. Pack this project and install it by running 
-``npm install $(npm pack ../frontend-build/ | tail -1)``. This should work every time but
-you will need to run it every time you make changes to this project.
+Because of this, local development can be tricky. The easiest way to do local
```
abutterworth: It's not nearly as tricky as it used to be

bin/fedx-scripts.js
```diff
@@ -1,19 +1,64 @@
...
+// commandName is the third argument after node and fedx-scripts
+const commandName = process.argv[2];
+
+// remove fedx-scripts from process.argv to allow subcommands to read options properly
+process.argv.splice(1, 1);
+
+switch (commandName) {
   case 'babel':
+    ensureConfigOption(presets.babel);
+    require('@babel/cli/bin/babel');
```
abutterworth: By simply requiring this script we're running it as if run by `npm run babel`

bin/fedx-scripts.js
```diff
@@ -1,19 +1,65 @@
 #!/usr/bin/env node
+const presets = require('../lib/presets');
 
-process.env.PROJECT_ROOT = process.cwd();
```
abutterworth: Removed this everywhere it's used and swapped for process.cwd() since we now have less trouble controlling which directory scripts are being executed in.

bin/fedx-scripts.js
```diff
@@ -1,19 +1,65 @@
...
+/**
+ * This file executes forwards cli commands by manipulating process.argv values
```
davidjoy: "executes forwards" 🤔

PR #113, opened 2020-09-24
fix: transition to use dart sass in webpack config

adamstankiewicz:
Transitions @edx/frontend-build to use dart sass over libsass. See https://github.com/edx/frontend-build/issues/111 for more details.


abutterworth (approved):
Is this a breaking change?

davidjoy (approved):
Approved - I'd like to sanity check that this "just works" when our dependents upgrade to it, but it looks fine otherwise.

adamstankiewicz:
> Is this a breaking change?

@abutterworth According to the [docs](https://www.npmjs.com/package/dart-sass), "Node Sass supports the same API" as dart-sass, so I don't believe it is a breaking change. If we were using Ruby Sass, there would be breaking changing but I am not finding anything that leads me to believe moving from node-sass (LibSass) to dart-sass is breaking.

PR #119, opened 2020-10-01
Add capability to configure publicPath of webpack through environmental variables.

morenol:
This PR adds the capability to configure the [publicPath](https://webpack.js.org/guides/public-path/) webpack configuration through environmental variables in order to allow the deployment of MFEs in custom paths.

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -2,6 +2,7 @@
 // optimized bundles at the expense of a longer build time.
 const { merge } = require('webpack-merge');
 const path = require('path');
+const dotenv = require('dotenv');
 const Dotenv = require('dotenv-webpack');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 const HtmlWebpackNewRelicPlugin = require('html-webpack-new-relic-plugin');
@@ -15,12 +16,18 @@ const CssNano = require('cssnano');
 const commonConfig = require('./webpack.common.config.js');
 const presets = require('../lib/presets');
 
+// Add process env vars. Currently used only for setting the PUBLIC_PATH.
+dotenv.config({
+  path: path.resolve(process.cwd(), '.env'),
+});
+
 module.exports = merge(commonConfig, {
   mode: 'production',
   devtool: 'source-map',
   output: {
     filename: '[name].[chunkhash].js',
     path: path.resolve(process.cwd(), 'dist'),
+    publicPath: process.env.PUBLIC_PATH || '/',
   },
   module: {
     // Specify file-by-file rules to Webpack. Some file-types need a particular kind of loader.
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -12,7 +12,8 @@ const PostCssRtlPlugin = require('postcss-rtl');
 const commonConfig = require('./webpack.common.config.js');
 const presets = require('../lib/presets');
 
-// Add process env vars. Currently used only for setting the server port
+// Add process env vars. Currently used only for setting the
+// server port and the publicPath
 dotenv.config({
   path: path.resolve(process.cwd(), '.env.development'),
 });
@@ -78,6 +79,9 @@ module.exports = merge(commonConfig, {
     // enable react's custom hot dev client so we get errors reported in the browser
     hot: require.resolve('react-dev-utils/webpackHotDevClient'),
     app: path.resolve(process.cwd(), 'src/index'),
+  },
+  output: {
+    publicPath: process.env.PUBLIC_PATH || '/',
   },
   resolve: {
     alias: aliases,
@@ -202,6 +206,6 @@ module.exports = merge(commonConfig, {
     historyApiFallback: true,
     hot: true,
     inline: true,
-    publicPath: '/',
+    publicPath: process.env.PUBLIC_PATH || '/',
   },
 });


edx-webhook:
Thanks for the pull request, @morenol! I've created [OSPR-5028](https://openedx.atlassian.net/browse/OSPR-5028) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

config/webpack.prod.config.js
```diff
@@ -167,5 +172,8 @@ module.exports = Merge.smart(commonConfig, {
       analyzerMode: 'static',
       openAnalyzer: false,
     }),
+    new webpack.DefinePlugin({
```
davidjoy, line 175: We use the DotEnv plugin further up to pull environment variables in via `.env` files.  Can we fold this into that?  I'm not sure exactly what happens if both DotEnv and DefinePlugin are used together, perhaps it's fine, but we should keep it standardized.

natabene:
@morenol Thank you for your contribution. Please let me know once it is ready for our review.

config/webpack.prod.config.js
```diff
@@ -15,12 +15,15 @@ const CssNano = require('cssnano');
 const commonConfig = require('./webpack.common.config.js');
 const presets = require('../lib/presets');
 
+const BASENAME = process.env.BASENAME || '/';
```
morenol, line 18: @davidjoy I just realized that this is not relying on dotenv.

Should I add something like this? 
```
dotenv.config({
  path: path.resolve(process.cwd(), '.env'),
});
```

Or is it ok if this config is not configurable through dotenv and instead through environmental variables? If that is the case, I think  that I can apply this change to the webpack.common.config.js instead

morenol: Any thoughts? @davidjoy

config/webpack.common.config.js
```diff
@@ -1,12 +1,14 @@
 const path = require('path');
 
+const BASENAME = process.env.BASENAME || '/';
```
adamstankiewicz, line 3: nit: would `PUBLIC_PATH` potentially be a more descriptive environment variable name here?

config/webpack.prod.config.js
adamstankiewicz: @morenol I think putting this in webpack.common.config.js and reading through environment variables makes sense.

config/webpack.common.config.js
morenol: In terms of webpack, that is probably more appropriate. But we also want to use this in react-router,  in [frontend-platform](https://github.com/edx/frontend-platform/pull/113). The configuration there is called `basename`. So, not sure if we should use one or the other.

morenol:
> @morenol Thank you for your contribution. Please let me know once it is ready for our review.

Hi @natabene, This is ready for review!

adamstankiewicz (approved):

adamstankiewicz:
@morenol Per your earlier comment, I think we'll end up wanting to read those process.env variables in webpack.common.config.js with something like this after all:

```
const dotenv = require('dotenv');

// Add process env vars. Currently used only for setting the PUBLIC_PATH.
dotenv.config({
  path: path.resolve(process.cwd(), '.env'),
});
```

I will merge this PR once that change in place 😄 After it's deployed, semantic-release will automatically post a comment on this PR with the NPM version/tag it was published to. Thanks for the contribution!

morenol:
@adamstankiewicz Got it, my only complaint about that is that we would not be using the .env.development to set that configuration so there would be discrepancies with the dev environment

natabene:
@marcotuts Is this a blended ticket by any chance?

morenol:
> @marcotuts Is this a blended ticket by any chance?

Hi @natabene, I am almost sure that this is not part of a blended project

morenol:
> @morenol Per your earlier comment, I think we'll end up wanting to read those process.env variables in webpack.common.config.js

Done!

morenol:
> If this works for your use case, we likely don't need to modify webpack.common.config.js to change `publicPath` on a per-environment basis using an environment variable.

I think that it would work, I have only worked with the profile, account and gradebook MFEs. They seems to not have those files but I think that it would work. 

I think that we can keep at least the ability to change the publicPath using environements variables so it can be used without modification by the tubular scripts.

Other option that I think that it would be OK, is to remove the changes from webpack.common.config.js, and instead override the publicPath in webpack.prod.config.js and webpack.dev.config.js using their respectives .env files

adamstankiewicz:
If a microfrontend is using @edx/frontend-build, but does not have webpack.dev.config.js and/or webpack.prod.config.js, it just means those repos are using the default config presets. Adding the file as shown above will force it to read from the custom config rather than the default.

As for making @edx/frontend-build compatible with tubular, I think that's fair. The approach to override `publicPath` in each of the config presets (e.g., webpack.dev.config.js) using the dotenv set up that already exists in those files sounds good 👍 We'll want to verify that overriding the default config with custom webpack config files defined in the consuming project still works as expected.

For example, if I have `PUBLIC_PATH` environment variable in .env.development, but also override `publicPath` with my own custom webpack.dev.config.js file, I would expect webpack to use the `publicPath` from the custom config file over the environment variable.

morenol:
@adamstankiewicz I made this test with the current version of this PR.


1) In the devstack, I went to the container of the frontend-app-learning MFE and there I ran this:

```
npm install @edx/frontend-build@git+https://github.com/edunext/frontend-build.git
```
The master branch of that fork has the same that the branch used in this PR.

After that, I added to the .env.development in the learning repo this:

 ```
PUBLIC_PATH='/custom-path'
```

I restarted that container, and once it was up I checked that everything was ok by running `curl localhost:2000` and noticing that the response was:

```htmll
<!doctype html>
<html lang="en-us">
  <head>
      <title>Course kx iiiasdadasd | edX</title>
      <meta charset="utf-8">
      <meta name="viewport" content="width=device-width, initial-scale=1.0">
      <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
  </head>
  <body>
    <div id="root"></div>
  <script src="/custom-path/app.js"></script><script src="/custom_and_weird_path/hot.js"></script></body>
</html>
```

Notice that the src of the app.js file is under the public path defined in the .env.development file.

Then in the root of that repo, I created a webpack.dev.config.js file with this content:
```js
// webpack.dev.config.js
const { getBaseConfig } = require('@edx/frontend-build');
const config = getBaseConfig('webpack-dev-server');

/* Custom config manipulations */
config.output.publicPath = '/another-path';

module.exports = config;
```
I checked again with curl and the response was:

```html
<!doctype html>
<html lang="en-us">
  <head>
      <title>Course kx iiiasdadasd | edX</title>
      <meta charset="utf-8">
      <meta name="viewport" content="width=device-width, initial-scale=1.0">
      <link rel="shortcut icon" href="/favicon.ico" type="image/x-icon" />
  </head>
  <body>
    <div id="root"></div>
  <script src="/another-path/app.js"></script><script src="/another-path/hot.js"></script></body>
</html>
``` 

So, it seems that is it overriding the behavior of the frontend-build file, so I think that everything is working as expected.

adamstankiewicz (changes requested):
I added a couple additional comments that should help wrap up this PR.

> Not sure, but maybe in the MFE, we would need to change the path of the favicon.ico in public/index.html to include the PUBLIC_PATH environment variable once we upgrade the frontend-build version used in them.

Yes, using the example app in this repo, I confirmed that you'll need to modify the path of the favicon.ico public/index.html in the MFE for the favicon.ico to properly appear when using a custom `PUBLIC_PATH`. The below will use the `publicPath` from the Webpack's config for the build:

```html
<link rel="shortcut icon" href="<%=webpackConfig.output.publicPath%>favicon.ico" type="image/x-icon" />
```

adamstankiewicz (approved):
Looks good! Thanks for putting this together :) 

This package is released to NPM via semantic-release, which follows Conventional Commits message formats. Before we merge, do you mind squashing the commits for this PR and renaming the resulting commit to be something like:

`feat(webpack): configure webpack's publicPath through environment variables`

This will ensure semantic-release is aware that it needs to deploy the new version to NPM.

edx-webhook:
@morenol 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/frontend-build&entry.867055334=https://github.com/edx/frontend-build/pull/119&entry.1484655318=https://github.com/morenol&entry.752974735=2020-10-01+17:46&entry.1917517419=2020-10-16+16:01&entry.2133058324=Yes) so we can improve your experience in the future.

PR #133, opened 2020-10-20
fix: ensure publicPath works with HtmlWebpackPlugin

davidjoy:
The default value is “`auto`” rather than “`/`“ like the other places we’re using PUBLIC_PATH.

Without this, HtmlWebpackPlugin doesn’t generate proper paths to where the assets are being served.  If PUBLIC_PATH is set to “foobar”, it will still look for main at /main.js, rather than /foobar/main.js


davidjoy:
Don't think this PR is necessary.  Closing.

PR #141, opened 2020-12-10
ci: switch to GHA, add commitlint to ci checks

bbaker6225:


bbaker6225:
example output:
```
$ commitlint-travis

⧗   input: Add commitlint travis

✖   subject may not be empty [subject-empty]

✖   type may not be empty [type-empty]
````

adamstankiewicz (approved):

PR #147, opened 2020-12-17
fix: Downgrading image-webpack-loader to fix linux arch bug in gifsicle

davidjoy:
There’s a documented issue in “imagemin-gifsicle” where a recent update (5.0.0) removed the binaries for linux distributions:

https://github.com/imagemin/gifsicle-bin/issues/113
https://github.com/tcoopman/image-webpack-loader/issues/233

This causes it to explode in CI via our direct dependency on image-webpack-loader.  The 7.0.0 version of image-webpack-loader relies on version 5 of gifsicle.  If we downgrade back to 6.0.0, we get gifsicle 4.0.1 instead, which still has the binaries.

I'm also noticing that the source for gifsicle is a bit confusing.  There's two projects here:

https://github.com/imagemin/imagemin-gifsicle
https://github.com/imagemin/gifsicle-bin

gifsicle-bin is the one that needs to be downgraded to 4.0.1 from 5.0.0.  imagemin-gifsicle also gets downgraded from 7.0.0 to 6.0.1 in the process, and appears to be the direct dependency of image-webpack-loader.

Confusing, but I think this will fix our pipelines.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4759,9 +4759,9 @@
       "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw=="
     },
     "arch": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/arch/-/arch-2.1.2.tgz",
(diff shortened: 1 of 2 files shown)


adamstankiewicz (approved):

mikix (approved):
Is there a way to mark this so it doesn't get automatically upgraded again? Ideally with a comment?

ciduarte (approved):
Thanks for this fix! 👍

PR #148, opened 2020-12-21
feat: allow private overrides of dev env vars

adamstankiewicz:
This PR adds support for the proposed feature as outlined in: https://github.com/edx/frontend-build/issues/140

You may now create a `.gitignored`'ed .env.private file in any repo that uses edx/frontend-build's Webpack configs to override environment variables defined in .env.development and .env.development-stage.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -136,6 +136,18 @@ Note that the "dir" and "dist" keys give you granular control over the shape of 
 
 This mechanism uses Webpack resolve aliases, as documented here: https://webpack.js.org/configuration/resolve/#resolvealias
 
+Override default .env.development environment variables with .env.private
+-------------------------------------------------------------------------
+
+In some situations, you may want to override development environment variables defined in .env.development
+with private environment variables that should never be checked into a repository. For example, a
+.env.development file may contain secrets for a third-party service (e.g., Algolia) that you'd like to use
+during development but want to ensure these secrets are not checked into Git.
+
+You may create a `.env.private` with any overrides of the environment settings configured in `.env.development`.
+
+**Note: .env.private should be added to your project's .gitignore so it does not get checked in.**
+
 Development
 -----------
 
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -136,6 +136,18 @@ Note that the "dir" and "dist" keys give you granular control over the shape of
...
+-------------------------------------------------------------------------
+
+In some situations, you may want to override development environment variables defined in .env.development
+with private environment variables that should never be checked into a repository. For example, a
+.env.development file may contain secrets for a third-party service (e.g., Algolia) that you'd like to use
+during development but want to ensure these secrets are not checked into Git.
+
+You may create a `.env.private` with any overrides of the environment settings configured in `.env.development`.
+
+**Note: .env.private should be added to your project's .gitignore so it does not get checked in.**
```
binodpant, line 149: ah since this is a lib, each MFE is going to have to .gitignore.. (no way to auto gitignore I guess,... okay)

config/webpack.dev-stage.config.js
```diff
@@ -11,12 +11,18 @@ const PostCssRtlPlugin = require('postcss-rtl');
...
 
 // Add process env vars. Currently used only for setting the server port
 dotenv.config({
   path: path.resolve(process.cwd(), '.env.development-stage'),
 });
 
+// Allow private/local overrides of env vars from .env.development for config settings
+// that you'd like to persist locally during development, without the risk of checking
+// in temporary modifications to .env.development.
+resolvePrivateEnvConfig('.env.private');
```
binodpant, line 24: do we need it here or  just in webpack.dev.config.js file?

README.rst
adamstankiewicz: Correct.

davidjoy (approved):

binodpant (approved):
barring that question on the stage file (not sure when we use that), if this worked for you locally I am good to go! thanks for doing this, it's been very annoying to have to remember to not check in my algolia overrides!

config/webpack.dev-stage.config.js
adamstankiewicz: Yeah, let's keep it in both configs so that .env.private will apply to either :)

adamstankiewicz:
> barring that question on the stage file (not sure when we use that), if this worked for you locally I am good to go! thanks for doing this, it's been very annoying to have to remember to not check in my algolia overrides!

Agreed!

PR #165, opened 2021-02-17
fix: improving module config parsing output

davidjoy:
- If there are errors, print the errors.
- If the file doesn’t exist, don’t do anything.
- If we’re successful, describe what we’ve done.


adamstankiewicz (approved):
LGTM! :shipit:

PR #177, opened 2021-06-08
feat: upgrade build ecosystem and webpack to v5

davidjoy:
# BREAKING CHANGE: 

- Repositories that update to this release must also update `@edx/paragon` to `v16.0.0` or later.
- Repositories that update to this release must update `purgecss-webpack-plugin` to `v4.0.3` or later. This necessitates replacing `whitelistPatterns` with `safelist` in the webpack configuration block for PurgeCSS.

# Warnings
- The `devServer` configuration in webpack.dev.config.js and webpack.dev-stage.config.js is now using webpack-dev-server v4.0.0-beta.3, which changed the syntax of the devServer configuration and removed some values we had set such as `inline` and `disableHostCheck`.  It's not entirely clear to the author of this PR the effect of these changes, but we didn't observe any issues with the removal of those two config options.  
- Furthermore, we recognize we're using a beta of webpack-dev-server, but also haven't observed any issues with that.  We should upgrade to a stable release as soon as one is available.
- Similarly, we're using a beta of resolve-url-loader (5.0.0-beta.1) which was necessary to resolve versioning issues related to PostCSS (more on this below).  We didn't observe any issues with this beta, but should upgrade to a stable release as soon as one is available.

# Inform

## PostCSS versioning issues

One of the difficulties in this upgrade was ensuring all packages that depend on `postcss` are all using version 8.x of that library.  It needs to be installed as a peer dependency.  We wanted to include that dependency in `frontend-build` so that we don't have to create dependencies on `postcss` in all our micro-frontends that depend on `frontend-build`.  This has the unfortunate side effect that if a micro-frontend has a dependency that itself depends on `postcss` 7 or less, that dependency is likely to be installed in `node_modules/postcss` and get used by all frontend-build's dependencies that require `postcss` version 8.x.  When this happens, an `npm install` will list out several dozen broken peer dependency warnings (packages looking for postcss 8 and not finding it), and both the dev server and production builds will fail to load properly.

## RTL/LTR Stylesheet Generation

This version of frontend-build also changes the PostCSS plugin we use to perform creation of right-to-left stylesheets (used for some languages when the `dir="rtl"` attribute is present on the `html` element of the page.)  Before this version we used `postcss-rtl`, and now use `postcss-rtlcss`.  The implementation of `postcss-rtlcss` differs in that it creates a single stylesheet with both RTL and LTR styles in it, rather than two separate stylesheets.  This will have the effect of _doubling_ the perceived stylesheet size for app.css.  Practically speaking, though, it's the same number of bytes as it was with `postcss-rtl`, just in one file instead of two.  The end result is the same, but may be unexpected.

# Other changes

- The webpack.dev-stage.config.js configuration has been updated to include a configurable `PUBLIC_PATH` like the prod and dev configurations.
- The example app now includes some elements which can be used to determine if the RTL stylesheet is generated correctly.  To use it, you can replace `dir="ltr"` with `dir="rtl"` on the `html` element of the document. 

# Package upgrades

- @svgr/webpack 5.3.1 -> 5.5.0
- autoprefixer 9.8.6 -> 10.2.6
- css-loader 3.6.0 -> 5.2.6
- cssnano 4.1.11 -> 5.0.6
- dotenv-webpack 4.0.0 -> 7.0.3
- html-webpack-plugin 4.5.2 -> 5.3.1
- image-webpack-loader 6.0.0 -> 7.0.1
- mini-css-extract-plugin 0.12.0 -> 1.6.0
- postcss-loader 3.0.0 -> 6.1.0
- resolve-url-loader 3.1.3 -> 5.0.0-beta.1
- style-loader 1.1.4 -> 2.0.0
- webpack-cli 3.3.12 -> 4.5.0
- webpack 4.44.2 -> 5.21.2
- webpack-dev-server 3.11.2 -> 4.0.0-beta.3


.github/workflows/release.yml
```diff
@@ -23,7 +26,22 @@ jobs:
...
+              'next',
+              {
+                name: 'beta',
+                prerelease: true
+              },
+              {
+                name: 'alpha',
+                prerelease: true
+              }
+            ]
```
adamstankiewicz, line 44: FWIW, these are already the default options for semantic-release per the [documentation](https://semantic-release.gitbook.io/semantic-release/usage/configuration#branches).

```
['+([0-9])?(.{+([0-9]),x}).x', 'master', 'next', 'next-major', {name: 'beta', prerelease: true}, {name: 'alpha', prerelease: true}]
```

No harm in being explicit, though.

config/webpack.dev.config.js
```diff
@@ -140,10 +141,12 @@ module.exports = merge(commonConfig, {
...
             loader: 'postcss-loader',
             options: {
-              plugins: () => [
-                PostCssRtlPlugin(),
-                PostCssAutoprefixerPlugin({ grid: true }),
-              ],
+              postcssOptions: {
+                plugins: [
+                  PostCssAutoprefixerPlugin({ grid: true }),
+                  PostCssRTLCSS(),
```
adamstankiewicz, line 147: [curious] Does the ordering of these 2 plugins matter here? The ordering is reversed from what it was. My guess is "no", but figured it'd be worth asking.

example/src/index.jsx
```diff
@@ -2,6 +2,12 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './App';
 
+// This line is to emulate what frontend-platform does when i18n initializes.
+// It's necessary because our stylesheet is generated with `[dir="ltr"]` as a prefix on all
+// direction-sensitive CSS classes.  Without this line, those classes wouldn't be applied to the
+// document.  See: https://github.com/edx/frontend-platform/blob/master/src/i18n/lib.js#L186
+global.document.getElementsByTagName('html')[0].setAttribute('dir', 'ltr');
```
adamstankiewicz, line 9: 🎉

config/webpack.dev-stage.config.js
```diff
@@ -23,6 +25,8 @@ dotenv.config({
 // in temporary modifications to .env.development.
 resolvePrivateEnvConfig('.env.private');
 
+const PUBLIC_PATH = process.env.PUBLIC_PATH || '/';
```
adamstankiewicz, line 28: Great catch to add the `PUBLIC_PATH` stuff to webpack.dev-stage.config.js!

example/src/__snapshots__/App.test.jsx.snap
```diff
@@ -57,5 +57,16 @@ exports[`Basic test should render 1`] = `
   <p>
     Photo by Louis Hansel @shotsoflouis on Unsplash
   </p>
+  <p>
+    Test variable:
```
adamstankiewicz, line 61: Should `TEST_VARIABLE` also be added to a .env.test file so that the environment variable gets picked up in the snapshot?

adamstankiewicz (approved):
Looks great! I am very excited about this upgrade. I left a couple clarifying questions and suggestions. We'll also need to rebase and resolve conflicts in the package*.json files prior to merge.

example/src/__snapshots__/App.test.jsx.snap
davidjoy: Doh!  Good idea.

.github/workflows/release.yml
davidjoy: If they're the defaults, I'd rather just remove the config... this whole file is, I think, from when you created it for this branch initially. :)  I haven't really touched it.  What do you think we can remove here, the whole `branches` section from line 32 on?

adamstankiewicz: I'm also in favor of removing config when it's already the defaults. In this case, right, it'd be the entire `branches` section :) 

I wonder if it's also worth running the `release.yml` workflow for the `next-major` branch (line additions 6-8 in this file). Not sure if we'll ever actually run a `next-major` branch but might be good to still map the workflow to the default `branches` config.

PR #179, opened 2021-06-23
feat: use NewRelicSourceMapPlugin to upload source maps to New Relic

adamstankiewicz:
Utilizes the already installed `NewRelicSourceMapPlugin` NPM package in webpack.prod.config.js to upload source maps to New Relic for production builds.

With source maps uploaded to New Relic, viewing the unminified stack trace for JS errors occurs automatically as opposed to needing to manually upload a source map file.


robrap (approved):
Looks good. Thanks.
Don’t think the update to package-lock is needed in this PR, but it won’t hurt.

adamstankiewicz:
> Looks good. Thanks.
> Don’t think the update to package-lock is needed in this PR, but it won’t hurt.

@robrap It isn't, but I was getting warnings locally about the `caniuse` browser list needing to be updated so figured I'd resolve that while I'm in here 😄

PR #191, opened 2021-07-18
feat: Fix/improve hot reloading 🚀, support SCSS exports again, and add local alias overrides to webpack.dev-stage.config.js

adamstankiewicz:
### Local aliases in `webpack.dev-stage.config.js`
Adds support for local module overrides via `module.config.js` when using `webpack.dev-stage.config.js`.

### SCSS exports

While upgrading to Webpack 5 in `frontend-app-learner-portal-enterprise`, our SCSS exports were no longer getting imported in JS after upgrading to frontend-build v7, which upgraded several packages related to CSS and styles. 

Prior to the Webpack 5 related upgrades, the following would return an object of exported SCSS variables:

```scss
:export {
  primary: $primary;
  // ...
}
```
```jsx
import colors from './colors.scss';
```

where `colors` has a value of `{ primary: /* hex code */ }`. However, without the following change, `colors` is an empty object.

To resolve, the following option was added to `css-loader`:

```js
modules: {
  compileType: 'icss',
},
```

### Hot reloading

Additionally, it seems hot module replacement broke with the Webpack 5 upgrade, where it correctly console.log messages about reloading modules on code changes but does not actually re-render the component or page as it once did. This was also observed in the `frontend-build` example app.

The resolution here was to remove the `hot` entry point in the webpack.dev(-stage).config.js files.

Beyond that, this PR also enables React Fast Refresh. With our current hot reloading implementation, most code changes result in a full page refresh, losing previous component states during the change, leading engineers to need to spend time replicating scenarios such as going through a UX flow before arriving at the same point in the app as before to see their  changes. With Fast Refresh, component state is persisted and components are simply re-rendered, eliminating the need for a full page refresh in more situations and without needing to repeat UX flows when making code changes.

https://www.npmjs.com/package/react-refresh
https://www.npmjs.com/package/@pmmmwh/react-refresh-webpack-plugin

Note: `react-refresh-webpack-plugin` is still noted as experimental; however, `create-react-app` has enabled Fast Refresh via these plugins by default so it should be safe enough for our needs 🚀


davidjoy (approved):
Nice work!

PR #196, opened 2021-08-19
fix: use webpack 5 compatible new relic source map plugin

adamstankiewicz:


binodpant (approved):
lgtm, as the fork's PR has already been verfieid

PR #200, opened 2021-09-30
feat: JavaScript-based environment configuration

davidjoy:
This adds a fairly simple mechanism to the webpack config that resolves a package named `env.config` to a local `env.config.js` file in the app.  If such a file doesn't exit, it falls back to an empty `env.config.js` file in frontend-build so that webpack doesn't break.

This is the groundwork required for us to use JS-based environment configuration.  This would have numerous benefits, most importantly, letting us use non-string values in our config (like booleans and numbers) as well as importing other resources to be used as config, such as alternate implementations of services in frontend-platform.


tuchfarber:
Fully support this plan! One thing I want to call out to make sure we account for it is the multiple builds of the same app for the masters learner portal. A quick glance at the code makes it seem like this would be fine (the tubular would just create a separate env.config.js each time?), but throwing this out there just in case :)

davidjoy:
Cool, good call out.  

I think that's what it should do, yes.  I fully expect to have to make modifications to tubular to get this working... Details are TBD, but I sort of envision adding env.config.js files to edx-internal's environment-specific subdirs and then modifying tubular to paste those into the repo before running `npm run build` for any given environment.

config/webpack.common.config.js
```diff
@@ -9,6 +9,12 @@ module.exports = {
     publicPath: '/',
   },
   resolve: {
+    alias: {
+      'env.config': path.resolve(process.cwd(), './env.config'),
+    },
+    fallback: {
+      'env.config': path.resolve(__dirname, './env.config.js'),
+    },
```
ghassanmas, line 17: I wonder if you set the `fallback` as follow,
```js
fallback: {
      'env.config': false
 },
```
Wouldn't this account for error if the file doens't exist and thus there would be neccessity for creating the empty js file?

davidjoy: Nice!  That seems to work - it returns an empty object.  I'm going to have to do something for jest, since it's not using this webpack fallback stuff, but this simplifies the webpack part a bunch. Thanks, great suggestion.

docs/0002-js-environment-config.md
```diff
@@ -0,0 +1,45 @@
+# JavaScript-based environment configuration
```
adamstankiewicz, line 1: Great writeup! 💯 🏆

docs/0002-js-environment-config.md
```diff
@@ -0,0 +1,45 @@
...
+
+```
+import config from 'env.config';
+```
+
+And it'll import the code from `env.config.js` if it's present, or the frontend-build version (which is an empty object `{}`) if it is not.
+
+# Follow-on work
+
+After getting this code committed to frontend-build, we'll want to consume it in frontend-platform.  This will probably take the form of a new "configuration service" interface with two implementations: the existing `process.env`-based implementation packaged into a class, and a new `env.config`-based implementation.  Cutting over from one to the other over time will have to be done in a backwards compatible way so that we can continue to support existing MFE builds.
```
adamstankiewicz, line 45: Good callout for future changes need to get this implemented 🚀

Is it worth also noting that build scripts (e.g., tubular) may also need modifications to support this?

example/.eslintrc.js
```diff
@@ -0,0 +1,3 @@
+const config = require('../config/.eslintrc.js');
```
adamstankiewicz, line 1: Is this .eslintrc.js file necessary if it's just using the default eslint config in `../config` without extending it? Similar question for the existing `.babelrc.js` file as well.

PR #217, opened 2021-12-07
feat!: build cleanup and remove dependency on html-webpack-new-relic-plugin

davidjoy:
BREAKING CHANGE: This PR performs some surgery on the code that loads our New Relic browser agent into our MFEs.  Because of that, we're making this a breaking change and reminding MFE application owners to _verify their app is still sending data to New Relic_ after updating.  We have tested these changes in the frontend-build/example app, but you're responsible for double-checking your app's integration.  

Also:

1. cleaning up a warning about HotModuleReplacementPlugin, which stated that the plugin doesn't need to be included explicitly if `hot: true` has been specified.
2.  bumping webpack-dev-server into a real release, instead of a beta.
3. Adding an .nvmrc file set to node 12.


lib/plugins/HtmlWebpackNewRelicPlugin.js
```diff
@@ -0,0 +1,73 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+*/
+
+// This class comes from https://github.com/victorrodrigues/html-webpack-new-relic-plugin.
+// We've copied it here rather than forking because this is the only place we use it, and having it
+// in a separate package isn't useful.
```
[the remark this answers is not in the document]
davidjoy: Adjusted the language. 👍🏻

lib/plugins/HtmlWebpackNewRelicPlugin.js
```diff
@@ -0,0 +1,73 @@
...
+    compiler.hooks.compilation.tap('HtmlWebpackNewRelicPlugin', (compilation) => {
+      HtmlWebpackPlugin.getHooks(compilation).alterAssetTagGroups.tap(
+        'HtmlWebpackNewRelicPlugin',
+        (data) => {
+          const newRelicScriptTag = HtmlWebpackPlugin.createHtmlTagObject(
+            'script',
+            { type: 'text/javascript' },
+            this.newrelicstring,
+          );
+          data.bodyTags.push(newRelicScriptTag);
```
adamstankiewicz, line 66: If possible, I think we might want to use `data.headTags` here to add this NR script to the `head` vs. the `body` per the recommend of NR ([source](https://docs.newrelic.com/docs/browser/new-relic-browser/page-load-timing-resources/instrumentation-browser-monitoring/#javascript-placement)).

lib/plugins/HtmlWebpackNewRelicPlugin.js
```diff
@@ -0,0 +1,74 @@
...
+    compiler.hooks.compilation.tap('HtmlWebpackNewRelicPlugin', (compilation) => {
+      HtmlWebpackPlugin.getHooks(compilation).alterAssetTagGroups.tap(
+        'HtmlWebpackNewRelicPlugin',
+        (data) => {
+          const newRelicScriptTag = HtmlWebpackPlugin.createHtmlTagObject(
+            'script',
+            { type: 'text/javascript' },
+            this.newRelicString,
+          );
+          data.headTags.push(newRelicScriptTag);
```
davidjoy, line 67: Note that we're now pushing this to the end of the head tags.

adamstankiewicz (approved):
LGTM, assuming we've tested this locally (maybe through the example app) to make sure data is for sure sent up to New Relic.

robrap:
Thanks. Did you consider bringing along any of the tests or documentation?

mikix (approved):

davidjoy:
@robrap Good point - I've pulled the tests over.  I don't think the documentation is useful now, since it's no longer an npm package.  I did, however, add a lengthy comment on how to update it.

davidjoy:
I've also verified the data gets sent to NR properly. 🥳

PR #223, opened 2022-01-11
fix: upgrade image-webpack-loader to 8.1.0

adamstankiewicz:
[This upstream PR](https://github.com/tcoopman/image-webpack-loader/pull/413) was merged and released for `image-webpack-loader` to upgrade `imagemin-webp` from v6 to v7 to remove its dependency on `trim-newlines` which has a "High" security vulnerability at its current version.

As a result, we can upgrade to this new version of `image-webpack-loader` to resolve our security dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2561,12 +2561,6 @@
       "resolved": "https://registry.npmjs.org/arr-union/-/arr-union-3.1.0.tgz",
       "integrity": "sha1-45sJrqne+Gao8gbiiK9jkZuuOcQ="
     },
-    "array-find-index": {
-      "version": "1.0.2",
-      "resolved": "https://registry.npmjs.org/array-find-index/-/array-find-index-1.0.2.tgz",
-      "integrity": "sha1-3wEKoSh+Fku9pvlyOwqWoexBh6E=",
-      "optional": true
-    },
     "array-flatten": {
       "version": "1.1.1",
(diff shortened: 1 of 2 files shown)


ghassanmas (approved):
Kudos to you that you for pushing the upstream fix!.

davidjoy (approved):

manny-m (approved):

macdiesel (approved):

regisb:
Should this PR be applied to the Maple release? If yes that would mean upgrading frontend-build in all MFEs, correct?

ghassanmas:
I guess if that is true, it would be nice if [edx-tools](https://github.com/openedx/edx-tools) would have a feature to update an npm package for MFEs, I anticipate something like:
The script gets as input the following; 
- The package name and its target version
- The target MFEs that needs to be updated  
- Branch name 
- commit message

Then the flow would be something like 
- Check out branch 
- Commit a change with {"PacakgeName":"targetVersion"} for all the targeted MFEs
- Open PR for the changes?.

It's just my speculation, if it's something common to update all MFEs then it makes sense to develop such a script. however I just wonder why ["^"](https://docs.npmjs.com/about-semantic-versioning#using-semantic-versioning-to-specify-update-types-your-package-can-accept) isn't used because it installs the latest patches by default, I guess there might a reason why, I am just wondering.

PR #233, opened 2022-02-24
fix: remove grid option from PostCssAutoprefixerPlugin

adamstankiewicz:
By default, `PostCssAutoprefixerPlugin` will not autoprefix `grid`-based properties, unless it is explicitly enabled, which we are doing in frontend-build's Webpack configs. As we no longer officially support older browsers like IE 11, we no longer have a need to autoprefix `grid` properties. More details about `grid` and autoprefixer may be found [here](https://github.com/postcss/autoprefixer#does-autoprefixer-polyfill-grid-layout-for-ie).

A new Paragon component (`SelectableBox`) utilizes the CSS `grid` properties, but when upgrading an MFE to the Paragon version with this new component, there are Webpack build warnings related to autoprefixer not supporting the `grid-auto-rows` property:

image

In order for the MFE to build properly on this Paragon version is to have `grid` autoprefix explicitly disabled as we no longer have a need to support older browsers like IE 11 anymore.

With this change, MFEs using the aforementioned version of Paragon will build without warnings again.

### DIFF
--- a/config/webpack.dev-stage.config.js
+++ b/config/webpack.dev-stage.config.js
@@ -81,7 +81,7 @@ module.exports = merge(commonConfig, {
             options: {
               postcssOptions: {
                 plugins: [
-                  PostCssAutoprefixerPlugin({ grid: true }),
+                  PostCssAutoprefixerPlugin(),
                   PostCssRTLCSS(),
                 ],
               },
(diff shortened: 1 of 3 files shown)


long74100 (approved):
lgtm

PR #252, opened 2022-06-24
Upgrade eslint to v8.x 

BilalQamar95:
**Ticket**
[42: Upgrade eslint to v8.x](https://github.com/openedx/frontend-wg/issues/42)

**Description:** 
- Upgraded `eslint` to v8.18 along with `@edx/eslint-config`
- Resolved linting issues post upgrade.


davidjoy:
It's a bit nebulous with a package like frontend-build, but I _think_ I might consider this a `feat:` PR, possibly with a breaking change as described in my previous comment.

BilalQamar95:
So I do see some(if not most) of our apps, depending on frontend-build, to have linting issues with this update.

**Details regarding breaking changes** 
- [migrating to v7](https://eslint.org/docs/latest/user-guide/migrating-to-7.0.0)
- [migrating to v8](https://eslint.org/docs/latest/user-guide/migrating-to-8.0.0)

davidjoy (approved):
I say we consider this a breaking change in frontend-build if it'll result in linting issues.  That way folks can be intentional about upgrading to it, rather than having it blow up their build.

So FYI, to do that with semantic-release, the commit message needs to include `BREAKING CHANGE:` and we should describe the breaking changes in as much detail as we can.  That'll cause semantic-release to bump `frontend-build`'s version number by a major increment.

If we squash commits and specify a new commit message with `feat:` in the subject and `BREAKING CHANGE:` in the body, that'll be sufficient.  I'm going to approve with the expectation that we'll just do that.

PR #253, opened 2022-07-20
Added configFile to fix linting issue

BilalQamar95:
**Ticket:** 
[@edx/frontend-build v12+ is not properly linting with eslint](https://github.com/openedx/frontend-wg/issues/114)

**What has changed:**
- Without configFile, linting was not working properly. Updated configFile in eslintrc to fix linting issue.

### DIFF
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -1,8 +1,13 @@
+const path = require('path');
+
 module.exports = {
   extends: '@edx/eslint-config',
   parser: '@babel/eslint-parser',
   parserOptions: {
-    requireConfigFile: false,
+    requireConfigFile: true,
+    babelOptions: {
+      configFile: path.resolve(__dirname, './babel.config.js'),
+    },
   },
   rules: {
     'import/no-extraneous-dependencies': [
--- a/config/.eslintrc.js
+++ b/config/.eslintrc.js
@@ -1,4 +1,4 @@
-const path = require('path');
+const { babel } = require('../lib/presets');
 
 module.exports = {
   extends: '@edx/eslint-config',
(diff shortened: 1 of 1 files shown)


config/.eslintrc.js
```diff
@@ -1,8 +1,13 @@
+const path = require('path');
+
 module.exports = {
   extends: '@edx/eslint-config',
   parser: '@babel/eslint-parser',
   parserOptions: {
-    requireConfigFile: false,
+    requireConfigFile: true,
+    babelOptions: {
+      configFile: path.resolve(__dirname, './babel.config.js'),
```
adamstankiewicz, line 9: From the Github [issue](https://github.com/openedx/frontend-wg/issues/114):

> However, by needing the .eslintrc.js file to point to a babel.config.js file might cause complications here, since consumers can override this babel configuration, in which case we'd want the babelOptions.configFile to point to the consumer's overridden Babel configuration, not the default babel.config.js file.

Is there a way that this Babel config file could utilize any extended/overridden `babel.config.js` file consumers happen to be using, and only fall back to the default `babel.config.js` as needed?

For example, if a consumer defines their own `babel.config.js` file (or other supported file name) at the root of their repository, `@edx/frontend-build` uses that one instead of the default `babel.config.js` specified here ([source](https://github.com/openedx/frontend-build/blob/master/lib/presets.js#L5-L15)).

My concern here is that if a consumer is using a custom `babel.config.js` file in their repo, we might not want to point to the default `babel.config.js` file in the default ESLint config since the consumer has extended/overridden it in some way, and could result in a discrepancy between what `babel.config.js` is used for ESLint vs. Babel itself.

adamstankiewicz (approved):

PR #259, opened 2022-08-30
Replaced `image-webpack-loader` with `image-minimizer-webpack-plugin`

BilalQamar95:
# Ticket
[107: frontend-build: High security vulnerability needs semver-regex update (via image-webpack-loader)](https://github.com/openedx/frontend-wg/issues/107)

# Description:
- Replaced `image-webpack-loader` with `image-minimizer-webpack-plugin`
- Switched from `imagemin` to `sharp` plugin


config/webpack.dev-stage.config.js
```diff
@@ -124,28 +125,25 @@ module.exports = merge(commonConfig, {
...
-            },
-          },
-        ],
+        loader: 'file-loader',
       },
     ],
   },
+  optimization: {
+    minimizer: [
+      '...',
```
adamstankiewicz, line 134: What is this for?

config/webpack.dev.config.js
```diff
@@ -124,26 +125,25 @@ module.exports = merge(commonConfig, {
...
+  optimization: {
+    minimizer: [
+      '...',
+      new ImageMinimizerPlugin({
+        minimizer: {
+          implementation: ImageMinimizerPlugin.sharpMinify,
+          options: {
+            encodeOptions: {
+              ...['png', 'jpeg', 'jpg'].reduce((accumulator, value) => (
+                { ...accumulator, [value]: { progressive: true, quality: 65 } }
```
adamstankiewicz, line 141: [clarification/curious] Is the `progressive: true` applicable to all file extensions? In the previous implementation, `progressive` is only relevant to `mozjpeg`.

package.json
```diff
@@ -57,7 +57,7 @@
     "file-loader": "6.2.0",
     "html-webpack-plugin": "5.5.0",
     "identity-obj-proxy": "3.0.0",
-    "image-webpack-loader": "8.1.0",
+    "image-minimizer-webpack-plugin": "^3.3.0",
```
adamstankiewicz, line 60: nitL: pin the dependencies.

config/webpack.dev.config.js
BilalQamar95: `jpeg` & `jpg` seems to direct to same options, `png` also has option for progressive (interlace) scan 
Screen Shot 2022-09-09 at 2 31 40 PM

config/webpack.dev-stage.config.js
BilalQamar95: I believe that would be to extend default minimizer. [Package setup details](https://webpack.js.org/plugins/image-minimizer-webpack-plugin/)

config/webpack.dev-stage.config.js
```diff
@@ -124,28 +125,25 @@ module.exports = merge(commonConfig, {
...
+      new ImageMinimizerPlugin({
+        implementation: ImageMinimizerPlugin.sharpMinify,
+        options: {
+          encodeOptions: {
+            ...['png', 'jpeg', 'jpg'].reduce((accumulator, value) => (
```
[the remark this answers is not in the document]
BilalQamar95: I do see gif as part of encode options
```
type SharpEncodeOptions = {
  avif?: import("sharp").AvifOptions | undefined;
  gif?: import("sharp").GifOptions | undefined;
  heif?: import("sharp").HeifOptions | undefined;
  jpeg?: import("sharp").JpegOptions | undefined;
  jpg?: import("sharp").JpegOptions | undefined;
  png?: import("sharp").PngOptions | undefined;
  webp?: import("sharp").WebpOptions | undefined;
};
```

**GifOptions (with defaults)**
```
/** Number of animation iterations, a value between 0 and 65535. Use 0 for infinite animation. (optional, default 0) */
loop?: number | undefined;
/** delay(s) between animation frames (in milliseconds), each value between 0 and 65535. (optional) */
delay?: number | number[] | undefined;
/** Maximum number of palette entries, including transparency, between 2 and 256 (optional, default 256) */
colours?: number | undefined;
/** Alternative Spelling of "colours". Maximum number of palette entries, including transparency, between 2 and 256 (optional, default 256) */
colors?: number | undefined;
/** Level of CPU effort to reduce file size, between 1 (fastest) and 10 (slowest) (optional, default 7) */
effort?: number | undefined;
/** Level of Floyd-Steinberg error diffusion, between 0 (least) and 1 (most) (optional, default 1.0) */
dither?: number | undefined;
```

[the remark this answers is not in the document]
jmbowman: @BilalQamar95 Am I reading correctly that we'd need to explicitly configure it to minify gif images?  We should probably do that since the existing implementation does so.

abdullahwaheed (approved):

abdullahwaheed (approved):

arbrandes:
Before we merge:

* Can we squash, please?
* Should we change the commit message to other than "refactor"?  I'm sure the consumer-facing API didn't change, but it feels like this should not be surprise-bundled into a future "fix", and instead warrants its own version number.  "feat"?

PR #261, opened 2022-10-03
chore: update linting rules to prevent warnings

Tj-Tracy:
When we updated eslint to v8, more rules were added by default. This turns off rules that we don't want. It also forces arrow function syntax for functional react components.


muselesscreator (approved):
Request comment on optional-nature of component definition, but beyond that, LGTM.

PR #258, opened 2022-08-24
refactor: migrated off babel-plugin-react-intl

BilalQamar95:
# Ticket
[120: Migrate off babel-plugin-react-intl in favor of babel-plugin-formatjs](https://github.com/openedx/frontend-wg/issues/120)

# What has changed
- Migrated to `babel-plugin-formatjs`

PR #286, opened 2022-12-21
fix: remove messagesDir for babel-react-intl 8

leangseu-edx:
The support for `messageDir` has been drop. https://github.com/formatjs/formatjs/commit/47ca556c1405eec7b7b2660275ae84a019112b2b

There should be a future plan use `babel-plugin-formatjs` instead because `babel-react-intl` has already been archived and moved.

### DIFF
--- a/config/babel-preserve-modules.config.js
+++ b/config/babel-preserve-modules.config.js
@@ -19,7 +19,6 @@ module.exports = {
         [
           'react-intl',
           {
-            messagesDir: './temp/babel-plugin-react-intl',
             moduleSourceName: '@edx/frontend-platform/i18n',
           },
         ],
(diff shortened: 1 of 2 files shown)


adamstankiewicz:
@leangseu-edx What is the consequence of removing `messagesDir`? How will our i18n process still function without it, if this is the only change?

For example, I believe the path given to `messagesDir` is referenced in frontend-app-learning's `Makefile` around defining the i18n-related `Make` targets ([source](https://github.com/openedx/frontend-app-learning/blob/080d31e93411a528bab76bb022058396d45135b4/Makefile#L8-L9)).

> There should be a future plan use babel-plugin-formatjs instead because babel-react-intl has already been archived and moved.

@leangseu-edx The migration from `babel-react-intl` to `babel-plugin-formatjs` is something we prioritized with FED-BOM and looks to still be in progress ([issue](https://github.com/openedx/frontend-wg/issues/120) and [PR](https://github.com/openedx/frontend-build/pull/258)).

PR #294, opened 2023-01-23
feat: Enable typescript for production builds

marlonkeating:
Client MFE's are beginning to move towards using typescript (see https://github.com/openedx/frontend-app-admin-portal/pull/957), and so we want to enable it for production.

## Checklist

- [X] npm run build succeeds

### DIFF
--- a/config/jest.config.js
+++ b/config/jest.config.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const fs = require('fs');
+const { jsWithTs: tsjPreset } = require('ts-jest/presets');
 
 const presets = require('../lib/presets');
 
@@ -33,11 +34,12 @@ module.exports = {
     '/node_modules/(?!@edx)',
   ],
   transform: {
-    '^.+\\.[t|j]sx?$': [
+    '^.+\\.jsx?$': [
       'babel-jest',
       {
         configFile: presets.babel.resolvedFilepath,
       },
     ],
+    ...tsjPreset.transform,
   },
 };
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -61,12 +61,12 @@ module.exports = merge(commonConfig, {
       // The babel-loader transforms newer ES2015+ syntax to older ES5 for older browsers.
       // Babel is configured with the .babelrc file at the root of the project.
       {
-        test: /\.(js|jsx)$/,
+        test: /\.(js|jsx|ts|tsx)$/,
         exclude: /node_modules\/(?!@edx)/,
         use: {
           loader: 'babel-loader',
           options: {
-            configFile: presets.babel.resolvedFilepath,
+            configFile: presets['babel-typescript'].resolvedFilepath,
           },
         },
       },
--- a/example/src/Image.tsx
+++ b/example/src/Image.tsx
@@ -1,0 +1,11 @@
+import React, { CSSProperties } from "react";
+
+type ImageProps = {
+  src: string;
+  alt?: string;
+  style?: CSSProperties;
+};
+
+export default function Image(props:ImageProps) {
+  return <img {...props} />;
+}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,27 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "declaration": true,
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "jsx": "react",
+    "lib": ["dom", "es6", "dom.iterable"],
+    "isolatedModules": true,
+    "module": "ES6",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": false,
+    "noImplicitThis": true,
+    "noUnusedParameters": true,
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
+    "strictFunctionTypes": false,
+    "target": "ES6",
+    "rootDir": "example",
+    "outDir": "dist"
+  },
+  "include": ["example/**/*"],
+  "exclude": ["example/dist/*", "example/node_modules/*'"]
+}
--- a/README.rst
+++ b/README.rst
@@ -129,6 +129,7 @@ An example module.config.js file looks like the following.  You can copy this in
      ],
    };
 
+
 Steps
 ~~~~~
 
@@ -156,6 +157,14 @@ during development but want to ensure these secrets are not checked into Git.
 You may create a `.env.private` with any overrides of the environment settings configured in `.env.development`.
 
 **Note: .env.private should be added to your project's .gitignore so it does not get checked in.**
+
+Local module configuration for TypeScript
+-----------------------------------------
+
+#. Copy tsconfig.json into the root of the module
+#. Set "rootDir" to the root of the source code folders
+#. Set "include" to wildcard patterns specifying the subdirectories/files under rootDir where source code can be found
+#. Include any wildcards under rootDir that should be excluded using "exclude"
 
 Development
 -----------
--- a/example/src/App.jsx
+++ b/example/src/App.jsx
@@ -1,4 +1,5 @@
 import config from 'env.config';
+import Image from './Image.tsx';
 import appleUrl, { ReactComponent as Apple } from './apple.svg';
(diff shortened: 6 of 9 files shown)


config/webpack.prod.config.js
```diff
@@ -61,12 +61,19 @@ module.exports = merge(commonConfig, {
...
           options: {
-            configFile: presets.babel.resolvedFilepath,
+            configFile: presets['babel-typescript'].resolvedFilepath,
+            // Caches result of loader to the filesystem. Future builds will attempt to read
+            // from the cache to avoid needing to run the expensive recompilation process
+            // on each run.
+            cacheDirectory: true,
+            plugins: [
+              require.resolve('react-refresh/babel'),
+            ],
```
adamstankiewicz, line 76: Are these relevant for the production build? 

The production build does not run on the Webpack Dev Server like webpack.dev.config.js does, so anything dealing with hot reloading (e.g., `react-refresh`, `cacheDirectory`) isn't really applicable to production builds, which is why they are not included in the existing webpack.prod.config.js file.

marlonkeating: Thanks for pointing that out, unneeded lines have been removed.

config/jest.config.js
```diff
@@ -33,11 +34,12 @@ module.exports = {
     '/node_modules/(?!@edx)',
   ],
   transform: {
-    '^.+\\.[t|j]sx?$': [
```
adamstankiewicz, line 36: [curious] Is `babel-jest` supposed to transform `.js` files or only `.jsx` files?

example/src/Image.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { CSSProperties } from "react";
+
+type ImageProps = {
+  src: string;
+  alt?: string;
+  style?: CSSProperties;
+};
+// eslint-disable-next-line react/function-component-definition
```
adamstankiewicz, line 8: nit: I think it's worth resolving this eslint error as opposed to disabling it.

tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "compilerOptions": {
+    "allowSyntheticDefaultImports": true,
+    "declaration": true,
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "jsx": "react",
+    "lib": ["dom", "es2019"],
+    "isolatedModules": true,
+    "module": "esnext",
```
adamstankiewicz, line 10: [curious] Was these config options largely copy/pasted from `@edx/paragon`'s example (looks like the config is identical)?

If so, I think we need to change the `tsconfig.json` file in `@edx/paragon`, too haha.

Reason being, I don't think we should be setting `module` or `target` to be "esnext":

> The special ESNext value refers to the highest version your version of TypeScript supports. This setting should be used with caution, since it doesn’t mean the same thing between different TypeScript versions and can make upgrades less predictable ([docs](https://www.typescriptlang.org/tsconfig#target))

While Open edX no longer needs to transpile down to ES5 to supporting legacy old browsers (e.g., IE11), we should be transpiling down to ES6. I'm also going to file an issue to this effect in the `@edx/paragon` repository.

tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
adamstankiewicz, line 1: [curious] It might be worthwhile to consider how we might be able to define a standard `tsconfig` in `@edx/frontend-build` such that:
* It may be shared with consuming applications so they don't need to re-define their own in the root of their repos, too.
* It may be extended/customized by consuming applications.

adamstankiewicz, line 1: [curious] Should we add any documentation to the `README` regarding TypeScript support and how to use it in MFEs (e.g., does each MFE need to define a `tsconfig.json` file)?

config/jest.config.js
marlonkeating: According to the prior config, only .jsx files, so I kept it that way.  That said, I don't see any .js/.ts unit tests, so it might be good to add some and make sure this configuration handles them.

adamstankiewicz: Yeah, I agree. Feel free to defer on this for another PR and/or file an issue since its unrelated to the core TypeScript changes here, but also cool to include it here while we're thinking about it :) Up to you haha.

tsconfig.json
adamstankiewicz: See https://github.com/sindresorhus/tsconfig as an example repo that hosts a shared TypeScript config that gets extended by consuming applications:

```json
{
	"extends": "@sindresorhus/tsconfig",
	"compilerOptions": {
		"outDir": "dist"
	}
}
```

tsconfig.json
marlonkeating: It was originally copy-pasted from @edx/paragon.  I've updated the references to point to the example folder.

tsconfig.json
marlonkeating: Good call, added to readme.rst

marlonkeating: I looked around for mechanisms that we might be able to export the tsconfig similar to how we export things like the config.js's, but there don't appear to be any built-in mechanisms yet (much to the chagrin of many developers).  The closest thing I could find was an external library https://github.com/c-vetter/tsconfig.js

marlonkeating: Started conversation on creating that repo: https://github.com/openedx/tcril-engineering/issues/643

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
+    "declaration": true,
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "jsx": "react",
+    "lib": ["dom", "es6", "es2019"],
```
adamstankiewicz, line 9: I don't think we need to keep `es2019` in here, since we're saying we really want to transpile down to ES6 here.

Also, one of the questions posed in the Paragon tsconfig [PR](https://github.com/openedx/paragon/pull/1962/files#r1088764652) is whether we need `dom.iterable`; it doesn't automatically get included via `dom` here. I believe we're going to opt to include it, but curious to hear your thoughts here.

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "isolatedModules": true,
+    "module": "ES6",
+    "moduleResolution": "node",
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": false,
```
adamstankiewicz, line 14: [clarification] Does `noImplicitAny=false` make it possible to use Paragon imports (e.g., components) that don't have types defined (in the package or from `@types/paragon`)?

Presumably, this is the config option that lets consuming applications treat TypeScript as opt-in versus requiring everything to be strictly-typed, correct?

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
+    "strictFunctionTypes": false,
+    "target": "ES6",
+    "rootDir": "example",
+    "outDir": "dist"
+  },
+  "include": ["example/**/*"],
+  "exclude": ["example/dist/*", "example/node/*'"]
```
adamstankiewicz, line 26: [clarification] Is this supposed to be `example/node_modules/*`? I don't think there is a directory at `example/node`?

tsconfig.json
marlonkeating: That's my understanding.  Imported components that don't include type definitions are treated as 'any', and with `noImplicitAny: true` this would generate errors.

adamstankiewicz: Got it, cool. Just wanted to sanity check my understanding of that option, too haha :)

adamstankiewicz (approved):

tsconfig.json
marlonkeating: Good catch!  Fixed.

tsconfig.json
marlonkeating: Went with dom.iterable, since it's not included with dom, and seems like something we'd use at some point if we're not already.

abdullahwaheed:
@marlonkeating this is breaking in the consumer MFEs. we should launch it separately or with breaking changes

https://github.com/openedx/frontend-app-payment/actions/runs/4264210961/jobs/7421980077

abdullahwaheed:
@marlonkeating could you please check the issues resulting with these changes, else we'll have to revert this PR

PR #312, opened 2023-03-08
Move ts-jest pkg from devDependency to dependency

Mashal-m:


abdullahwaheed:
we have removed ts-jest in #315 so closing this

PR #315, opened 2023-03-13
Revert PR of Enable typescript for production builds

Mashal-m:
I have reverted this [PR](https://github.com/openedx/frontend-build/pull/294) because it was breaking consumer MFEs

PR #328, opened 2023-04-05
feat: Re-enable typescript for production builds

marlonkeating:
This is a fixed version of https://github.com/openedx/frontend-build/pull/294 (reverted in https://github.com/openedx/frontend-build/pull/315), with package.json that includes typescript libraries in dependencies instead of devDependencies, which ensures that the requires typescript libraries will get pulled into any consumers of openedx/frontend-build.

### Example App
- [x] npm run build succeeds 
- [x] npm run test succeeds
### Local MFE
using typescript pulling frontend-build but no typescript libraries
- [x] npm run build succeeds 
- [x] npm run test succeeds


abdullahwaheed:
Thank you @marlonkeating for updating this and fixing the issues.  Since now we have separate repo for [typescript-config](https://github.com/openedx/typescript-config), i think we should move all the typescript related configurations in that repo and just update the respective changes here.
cc: @adamstankiewicz

example/src/App.jsx
```diff
@@ -1,4 +1,5 @@
 import config from 'env.config';
+import Image from './Image.tsx';
```
adamstankiewicz, line 2: Shouldn't need to file extensions for ESLint.

example/src/Image.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+type ImageProps = {
+  src: string;
+  alt?: string;
+  style?: CSSProperties;
+};
+
+export default function Image(props:ImageProps) {
+  return <img {...props} />;
+}
```
adamstankiewicz, line 11: nit: newline

tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+      "sourceMap": true,
+      "strict": true,
+      "strictFunctionTypes": false,
+      "target": "ES6",
+      "rootDir": "example",
+      "outDir": "dist"
+    },
+    "include": ["example/**/*"],
+    "exclude": ["example/dist/*", "example/node_modules/*'"]
+  }
```
adamstankiewicz, line 27: nit: newline

adamstankiewicz (commented):
CI is failing on `npm run lint` due to omitted `.tsx` file extensions. I believe we need to modify our ESLint config to add TypeScript support so it doesn't complain about things like this.

This can be achieved with `airbnb-typescript`, but would recommend adding it to our shared `@edx/eslint-config` library, given that's where our ESLint config extends `airbnb` and `airbnb/hooks` already.

See this PR: https://github.com/openedx/eslint-config/pull/124

package.json
```diff
@@ -73,6 +77,8 @@
     "sharp": "^0.31.0",
     "source-map-loader": "^4.0.1",
     "style-loader": "3.3.2",
+    "ts-jest": "^26.5.0",
+    "typescript": "^4.9.4",
```
adamstankiewicz, line 81: [clarification] Should `typescript` be removed from the `devDependencies` given its now installed in `dependencies`?

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+  "config/.eslintrc.js",
+  "example/.eslintrc.js",
+  "example/**/*",
+  "lib",
+  "bin",
+  "config"],
+  "exclude": ["node_modules",
+  "dist",
+  "example/dist/*",
+  "example/node_modules/*'"]
```
adamstankiewicz, line 17: nit: indenting (or lack thereof) makes this difficult to read at first glance:

```js
{
  "include": [
    ".eslintrc.js",
    "config/.eslintrc.js",
    "example/.eslintrc.js",
    // ...
  ]
}
```

example/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+  "include": [
+    ".eslintrc.js",
+    "env.config.js",
+    "src"
+  ],
+  "exclude": [
+    "node_modules",
+    "dist",
+  ]
+}
```
adamstankiewicz, line 16: nit: newline needed

example/src/Image.tsx
```diff
@@ -0,0 +1,16 @@
...
+};
+
+const Image = ({ alt, ...rest }:ImageProps) => <img alt={alt} {...rest} />;
+
+const defaultProps = {
+  alt: undefined,
+  style: undefined,
+};
+Image.defaultProps = defaultProps;
+export default Image;
```
adamstankiewicz, line 16: nit: newline needed

adamstankiewicz (approved):
LGTM to merge into `alpha`, though with a clarifying question first about the `typescript` dependency and a few nits.

PR #365, opened 2023-05-20
feat: expose `PARAGON` as a global variable

adamstankiewicz:
https://github.com/openedx/paragon/issues/2321

In [this post](https://github.com/openedx/frontend-platform/pull/440#discussion_r1189917137), @xitij2000 had a great suggestion to support version substitution on the Paragon theme urls such that if the theme url specified in configuration looks something like, `https://cdn.jsdelivr.net/npm/@edx/paragon@$paragonVersion/dist/core.css`, we could replace the `$paragonVersion` with the version of the locally installed Paragon in the application itself.

This approach should help mitigate the concern around the coupling between the Paragon version installed in `package.json` in consuming MFEs (related to [this PR](https://github.com/openedx/frontend-platform/pull/440)) as, if the theme URLs contain `$paragonVersion`, consumers won't need to worry about updating Paragon versions in two places, or using an incompatible version of the Paragon theme between the version specified in configuration and the locally installed Paragon version.

## Output

If there is no compatible version of `@edx/paragon` installed (i.e., does not include a `paragon-theme.json` file in its `dist` directory in `node_modules`), falls back to create an `undefined` global variable, which prevents consuming MFE runtimes from breaking due to usage of an undefined `PARAGON` global variable:

```js
const PARAGON_THEME = undefined;
```

_Consuming libraries (e.g., `@edx/frontend-platform`) should not expect `PARAGON_THEME` to always be defined._

### Development build (`npm start`)

image

### Production build (`npm run build`)

Note: includes the `[contentHash]` due to options in `webpack.config.prod.js`. When the Paragon version changes, we will invalidate the cached file `paragon-theme-{url}.{hash}.js` due to the `{hash}` changing when the files themselves change.

image


example/src/__snapshots__/App.test.jsx.snap
```diff
@@ -104,5 +104,11 @@ exports[`Basic test should render 1`] = `
   >
     I'm aligned right, but left in RTL.
   </p>
+  <h2>
+    Exposed Paragon version
+  </h2>
+  <p>
+    "\\"20.39.1\\""
+  </p>
```
xitij2000, line 112: I think this snapshot is going to break quite often. Might make sense to keep it static for tests.

adamstankiewicz: Yeah, I agree haha. Will need to find another way to test.

config/webpack.common.config.js
```diff
@@ -19,4 +27,17 @@ module.exports = {
     },
     extensions: ['.js', '.jsx'],
   },
+  plugins: [
+    new webpack.DefinePlugin({
```
adamstankiewicz, line 31: Expose these variables as global variables usable by consuming MFEs and JS libraries.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,184 @@
...
+const {
+  getParagonVersion,
+  getParagonThemeCss,
+  replacePeriodsWithHyphens,
+} = require('../../../config/data/paragonUtils');
+
+const paragonVersion = getParagonVersion(process.cwd());
+const paragonThemeCss = getParagonThemeCss(process.cwd());
+
+class ParagonWebpackPlugin {
```
davidjoy, line 12: Might want a class comment here that just describes what this plugin's job is.  I'm assuming it's adding a script with the `PARAGON` variable in it, but it's a lot of fairly dense webpack plugin code and I'm not sure if I'm missing something. :)

example/package.json
```diff
@@ -8,16 +8,19 @@
...
+    "start": "../bin/fedx-scripts.js webpack-dev-server",
+    "serve": "npx serve dist"
   },
   "keywords": [],
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0"
+    "@edx/paragon": "21.0.0-alpha.30",
```
davidjoy, line 18: Reminder to update meeeeee.

config/webpack.common.config.js
```diff
@@ -19,4 +30,16 @@ module.exports = {
     },
     extensions: ['.js', '.jsx'],
   },
+  optimization: {
```
davidjoy, line 33: I see the various `chunks: ['app']` below, and this optimization block here too.  What's the net effect of how we've changed the chunking?  What do the cacheGroups do for us?

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,230 @@
...
+      themeVariantEntryNames: {},
+    };
+
+    if (!paragonThemeCss && !brandThemeCss) {
+      return;
+    }
+
+    if (paragonThemeCss) {
+      // Core Paragon
+      this.paragon.coreEntryName = replacePeriodsWithHyphens(paragonThemeCss.core.entryName);
```
viktorrusakov, line 35: Doesn't this return the same hyphened string that already exists inside `paragonThemeCss.core` under `outputChunkName` property?
I think all usages of `replacePeriodsWithHyphens` can be removed from this file in favour of using `outputChunkName` property returned by `getParagonThemeCss` function unless I'm missing something.

Also, if you do that I feel like you can either significantly simplify `construct` method or just remove it altogether because `paragonThemeCss` and `brandThemeCss` already contain all of the data you put into `paragon` and `brand` objects, so you can just use them in the class instead.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,230 @@
...
+        this.brand.themeVariantEntryNames[key] = {
+          entryName: replacePeriodsWithHyphens(value.entryName),
+          default: value.default,
+          dark: value.dark,
+        };
+      });
+    }
+  }
+
+  logger(message) {
```
viktorrusakov, line 58: I don't think this method is used anywhere and also it probably won't work because `console` is a browser API and webpack does not have access to it.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,230 @@
...
+            coreCssAsset: paragonCoreCssAsset,
+            themeVariantCssAssets: paragonThemeVariantCssAssets,
+          } = this.getParagonCssAssetsFromCompilation(compilation);
+
+          const {
+            coreCssAsset: brandCoreCssAsset,
+            themeVariantCssAssets: brandThemeVariantCssAssets,
+          } = this.getParagonCssAssetsFromCompilation(compilation, { isBrandOverride: true });
+
+          let scriptContents;
```
viktorrusakov, line 187: Do you think it would be simpler to just define `const scriptContents = {}` and remove `createEmptyScriptContentsIfUndefined` function?

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,230 @@
...
+              },
+            };
+          }
+
+          const newSource = this.insertParagonScriptIntoDocument({
+            originalSource,
+            coreCssAsset: paragonCoreCssAsset,
+            themeVariantCssAssets: paragonThemeVariantCssAssets,
+            scriptContents,
+          });
```
viktorrusakov, line 222: I think `insertParagonScriptIntoDocument` only uses `originalSource` and `scriptContents` variables, so you can remove others.

config/webpack.common.config.js
adamstankiewicz: Correct, the chunking is staying the same as is it today, with the addition of splitting out the "locally" installed already-compiled CSS files from `@edx/paragon` and `@edx/brand`. I will add a comment :)

That said, we probably should find some ways to optimize our MFE Webpack bundles (e.g., test hypothesis that many small JS files for the `node_modules` is better than 1 larger JS file, recommend/implement a strategy for code splitting for MFEs): https://github.com/openedx/wg-frontend/issues/173

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
adamstankiewicz: Correct, `logger` is no longer used anywhere (was helpful to have around for debugging during development). FWIW, though, the `console.log` works during `npm run build` and `npm run start`, running in the terminal through Node.js.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
adamstankiewicz: Ah, yup. Good catch! Left over from a prior iteration of the implementation. Will remove.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
adamstankiewicz: D'oh, it would seem that way! Just a tad bit over-engineered it seems 😜

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
viktorrusakov: Yeah, I think I had a brain lag moment when I wrote this comment about `console.log`, I've used it the same way thousand times 🤦 It does exist in Node.js.

adamstankiewicz: I'm likely going to keep it around due to its helpfulness for debugging/adding temporary logs haha. Just used it to debug a few things to respond to your other comments 😅

viktorrusakov: Now that we confirmed that it works, I don't mind 😅

Mashal-m:
Hey @adamstankiewicz, What is the current status of this PR, is it ready to review and merge? 
Could you please resolve conflicts?

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * @returns {Object} Metadata about the Paragon and brand theme URLs from configuration.
+   */
+  async resolveParagonThemeUrlsFromConfig() {
+    // const mfeRuntimeApiConfig = await getMfeRuntimeConfig();
+    let envConfig;
+    if (fs.existsSync(envConfigFile)) {
+      envConfig = require(envConfigFile);
+    }
+    // const paragonThemeUrls = mfeRuntimeApiConfig?.PARAGON_THEME_URLS
```
01001110J, line 72: Hello, I would suggest removing these comments if they are not intended to be used in the future.

lib/plugins/paragon-webpack-plugin/utils.js
```diff
@@ -0,0 +1,388 @@
...
+  const {
+    APP_ID,
+    MFE_CONFIG_API_URL,
+  } = process.env;
+  const queryParams = new URLSearchParams({
+    mfe: APP_ID,
+  });
+  const url = `${MFE_CONFIG_API_URL}?${queryParams.toString()}`;
+  const response = await axios.get(url);
+  return response.data;
```
01001110J, line 15: Could we consider incorporating a try/catch block for this operation? It would be beneficial to anticipate and handle potential failures in the request.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,156 @@
...
+  getParagonStylesheetUrls,
+  injectParagonCoreStylesheets,
+  injectParagonThemeVariantStylesheets,
+} = require('./utils');
+const resolvePrivateEnvConfig = require('../../resolvePrivateEnvConfig');
+
+// Add process env vars. Currently used only for setting the
+// server port and the publicPath
+dotenv.config({
+  path: path.resolve(process.cwd(), '.env.development'),
```
andrey-canon, line 21: This makes that the build(`npm run build`) process uses env.development setttings instead of the env

example/env.config.js
```diff
@@ -3,4 +3,29 @@ module.exports = {
   CORRECT_BOOL_VALUE: 'Good, false meant false.  We did not cast a boolean to a string.',
   INCORRECT_BOOL_VALUE: 'Why was a false boolean true?',
   INTEGER_VALUE: 123,
+  PARAGON_THEME_URLS: {
+    core: {
+      urls: {
+        default: 'https://cdn.jsdelivr.net/npm/@openedx/paragon@alphaa/dist/core.min.css',
```
dcoa, line 9:
```suggestion
        default: 'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/core.min.css',
```

davidjoy:
@adamstankiewicz may have a more comprehensive view of this than I do, as I'm trying to page this paragon work back in to my brain.  But my sense of it (including openedx/frontend-platform#440) is that:

1. Sharing a stylesheet between MFEs becomes our default behavior once we have a 'shell' app which loads all the other MFEs, so nothing special will be necessary for that.
2. Dark mode / themes is still something we want, though we don't actually have a 'dark' theme yet, right?  So this is about groundwork for that, which may be easier to reason when the dust starts settling on OEP-65.

adamstankiewicz:
> 1. Sharing a stylesheet between MFEs becomes our default behavior once we have a 'shell' app which loads all the other MFEs, so nothing special will be necessary for that.

@davidjoy When you say "nothing special will be necessary" for sharing a stylesheet between MFEs as it will become the default behavior once we have a "shell" app, are you referring to the changes in the https://github.com/openedx/frontend-platform/pull/440 no longer being necessary? That is, the frontend-platform PR enables support for using the Paragon CSS from a CDN URL instead of importing directly from the `@openedx/paragon` NPM package (i.e., better supports multi-tenancy use cases). If we want to continue relying on external CDN URLs (versus importing from `node_modules`), then I imagine there would work to be done in the shell to embed the Paragon CSS CDN url.

> Dark mode / themes is still something we want, though we don't actually have a 'dark' theme yet, right? So this is about groundwork for that, which may be easier to reason when the dust starts settling on OEP-65.

Correct, there is no official dark mode (yet). The design tokens work in Paragon sets up the foundational ground work to support a dark mode in the future (and maybe a high-contrast mode even longer term) and the frontend-build/platform PRs try to support those future use cases. The actual theme switching capabilities is not a strict requirement for the design tokens initial release, though, but would like to ensure the technical design for consuming the updated Paragon CSS from design tokens can be readily extended to support themes in the future.

> do you feel this (external CSS stylesheets) is something we should push before, in parallel, as part of, or after OEP-65?

@arbrandes @davidjoy Supporting the external CSS stylesheets (i.e., loading Paragon's CSS via CDN url vs. importing from `node_modules`) proposed by this PR and the associated frontend-platform [PR](https://github.com/openedx/frontend-platform/pull/440) is probably not as relevant to OEP-65. In the short term, the "shell" app could still rely on importing the Paragon CSS via `node_modules`; however, I believe Open Craft / eduNext / etc. does desire the ability to load the Paragon CSS from external CDN URL for multi-tenancy support (i.e., not requiring apps to rebuild & redeploy whenever the theme changes).

Likely the biggest overlap with OEP-65 and Paragon design tokens is not so much whether we adopt the external CSS stylesheets proposed by these 2 PRs, but more so Paragon migrating from SCSS -> CSS variables. That is, specifically, around the current styles challenge where custom SCSS files in MFEs using Paragon's SCSS variables need to either be imported by the MFE's styles entry point (e.g., `src/index.scss`) or the custom SCSS files would need to re-import Paragon styles (risking duplicate Paragon styles).

This current challenge around needing to import Paragon SCSS variables before using the variables in custom SCSS files will be a non-issue after the SCSS -> CSS migration as part of Paragon design tokens. That is, custom SCSS will be able to reference Paragon's CSS variables without needing to import anything from Paragon into the custom SCSS file(s) first.

When a "guest" MFE has custom styles relying on Paragon's variables, but Paragon's CSS is provided by the "shell", we would want to be sure we do not risk introducing duplicate Paragon styles, as some MFEs are doing today. As it relates to OEP-65 and its shared CSS in the "shell", we could probably start by importing Paragon's CSS from `node_modules` in the "shell" but may likely need the CSS variables support from design tokens to prevent duplicate Paragon CSS introduced by "guest" apps due to the existing SCSS import issue described above.

Hope this helps!

davidjoy:
Thanks Adam!

> When you say "nothing special will be necessary" for sharing a stylesheet between MFEs as it will become the default behavior once we have a "shell" app, are you referring to the changes in the https://github.com/openedx/frontend-platform/pull/440 no longer being necessary?

I'm just saying that if the shell loads the brand and paragon CSS, then all the other MFEs loaded into it don't need to... assuming they want a compatible version.  This was under the assumption that the external CDN URL was only about downloading the stylesheet once; sounds like there's more to it than that.

adamstankiewicz:
Closed in favor of https://github.com/openedx/frontend-build/pull/546

PR #404, opened 2023-06-24
feat: add fedx-scripts serve

adamstankiewicz:
# Description

These changes propose adding a `serve` command to the `fedx-scripts` CLI such that consumers can create an NPM script to serve the production Webpack build from the generated `dist` directory when running `npm run build`. By serving the production build locally, engineers can have more confidence in their resulting production builds before it reaches a staging/production environment.

As is, there is no documented way to preview a production build of micro-frontends of the files in the `dist` directory. Simply opening the `index.html` does not work as the React routing breaks, amongst other issues.

The proposed `serve` command starts an [Express.js](https://expressjs.com/) serve running on the `PORT` setting defined in `env.config.js`. Note: the `.env*` files to configuration is now [deprecated](https://github.com/openedx/frontend-platform/blob/master/docs/decisions/0007-javascript-file-configuration.rst#decision) so the `serve` command is intentionally not supporting it.

[ADR](https://github.com/openedx/frontend-build/blob/deeac588e73733c7d9b9bfaf3b458ae7b473c303/docs/0003-fedx-scripts-serve.md)

[README documentation](https://github.com/openedx/frontend-build/blob/c756d237c9f5b9314aeaae01f39ac6b772f6808b/README.rst#serving-a-production-webpack-build-locally)


brobro10000 (approved):
Sad to see that the `.env*` isn't supported, but maybe the catalyst needed to start transitioning to `env.config.js` across other MFEs
👍🏽

bin/fedx-scripts.js
```diff
@@ -1,4 +1,7 @@
 #!/usr/bin/env node
+
+const chalk = require('chalk');
```
davidjoy, line 3: Love me some chalk. 😋

package.json
```diff
@@ -55,23 +56,24 @@
     "eslint-plugin-jsx-a11y": "6.7.1",
     "eslint-plugin-react": "7.32.2",
     "eslint-plugin-react-hooks": "4.6.0",
+    "express": "^4.18.2",
```
davidjoy, line 59:
```suggestion
    "express": "4.18.2",
```

davidjoy (dismissed):
Love it.  :shipit:

adamstankiewicz:
@davidjoy I've updated this PR to add support for either `.env.development|private` file(s) _or_ `env.config.js` file. Please take another look when you get a chance!

package.json
brobro10000, line 59: Bump on pinning this version 👍🏽

lib/scripts/serve.js
```diff
@@ -0,0 +1,76 @@
...
+// Add process env vars. Currently used only for setting the
+// server port and the publicPath
+dotenv.config({
+  path: path.resolve(process.cwd(), '.env.development'),
+});
+
+// Allow private/local overrides of env vars from .env.development for config settings
+// that you'd like to persist locally during development, without the risk of checking
+// in temporary modifications to .env.development.
+resolvePrivateEnvConfig('.env.private');
```
brobro10000, line 18: 🥳 , but good stepping stone to transition over to `env.config.js`

brobro10000 (approved):
👍🏽

brobro10000 (approved):
Third times the charm!

PR #429, opened 2023-07-28
Alpha

adamstankiewicz:


.github/workflows/commitlint.yml
```diff
@@ -7,4 +7,14 @@ on:
...
-    uses: openedx/.github/.github/workflows/commitlint.yml@master
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: remove tsconfig.json # see issue https://github.com/conventional-changelog/commitlint/issues/3256
+        run: |
+          rm tsconfig.json
```
arbrandes, line 18: We're not going to need this, going forward.  See this comment on a Paragon PR: https://github.com/openedx/paragon/pull/2867#discussion_r1406343198

example/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../tsconfig.json",
```
arbrandes, line 2: This doesn't match the README.  Should we change it?

README.md
```diff
@@ -218,6 +218,25 @@ locally. To serve a production build locally:
...
+## Local module configuration for TypeScript
+
+#. Create file in repository `tsconfig.json`, with a clause `"extends": "@openedx/frontend-build"`
+#. Set "rootDir" to the root of the source code folders
+#. Set "include" to wildcard patterns specifying the subdirectories/files under rootDir where source code can be found
+#. Include any wildcards under rootDir that should be excluded using "exclude"
+
+```Sample json
+{
+  "extends": "@openedx/frontend-build",
```
arbrandes, line 230: It looks like we're actually extending `@edx/typescript-config` in the tsconfig in this repo.  Should we not recommend doing the same elsewhere?

arbrandes (changes requested):
No objections in principle, but I feel we should answer a couple of questions.

arbrandes:
Also, we need an interactive rebase on master here to consolidate the commits into a couple of meaningful ones before merging.  I suggest:

* One for all the Typescript stuff
* One for the Jest Upgrade
* One for the Renovate stuff package-lock

I can try and do that if there are no objections.  It will require a force-push to the branch, though, so... heads up.

arbrandes:
And for the future, `alpha` should always be treated like an integration branch (because it is).  That means that we shouldn't take PRs to it directly.  Instead, we should have feature branches based on master (aka PRs) that get rebased often and then octopus-merged into `alpha` periodically.  That way it becomes relatively easy to decide to abandon a feature, or pull another one in.

Also, `alpha` is meant to be alpha-quality.  No guarantees, and expect breaking changes.  Only use it temporarily, for testing purposes.

README.md
bradenmacdonald: Also, it might be good to recommend setting `outDir`, because I found in https://github.com/openedx/paragon/pull/3016 that if you extend `@edx/typescript-config` but don't set `outDir`, when you run `tsc` the files will get output to `./node_modules/@edx/typescript-config/dist/` - which is not that different from sending them to `/dev/null`. Most people won't be running `tsc` (rather webpack, babel, etc.) but if they do this heads up will save them some time trying to figure out why the output is not showing up.

adamstankiewicz: > It looks like we're actually extending @edx/typescript-config in the tsconfig in this repo. Should we not recommend doing the same elsewhere?

@arbrandes I believe the original intent for recommending `@openedx/frontend-build` here was that there might be MFE-specific additions/updates we want in the base TS config available to MFEs that don't make sense in the upstream `@edx/typescript-config` (which, in theory, could be used for more beyond just MFEs).

For example, the base ESLint config provided to MFEs via frontend-build extends `@edx/eslint-config`, but slightly differs. This approach to extend `@openedx/frontend-build` was largely following that paradigm.

That said, if supporting MFE-only TypeScript config isn't necessary (which is probably isn't; I don't really see legacy UI repos like edx-platform adopting TypeScript anytime soon), I agree recommending to extend `@edx/typescript-config` directly makes more sense.

bradenmacdonald: Ah, I've always wondered why `typescript-config` and `eslint-config` existed. Personally, I'd prefer to just consolidate them into `frontend-build` or `frontend-platform`.

abdullahwaheed:
@arbrandes suggested changes are done, could you please review it again

arbrandes:
Taking another look!

PR #435, opened 2023-08-08
fix: make ts-jest a dependency to try and ensure it makes it to consu…

muselesscreator:
…mers


package.json
```diff
@@ -85,6 +85,7 @@
     "sharp": "^0.32.0",
     "source-map-loader": "^4.0.1",
     "style-loader": "3.3.3",
+    "ts-jest": "^26.5.0",
     "typescript": "^4.9.4",
```
adamstankiewicz, line 89: consider: Let's ensure all these `dependencies` are pinned to explicit version like the others.

adamstankiewicz (approved):

package.json
muselesscreator: fair

PR #474, opened 2023-10-24
[do not merge] experimental alpha -> master

marlonkeating:
Experimental fixes for https://github.com/openedx/frontend-build/pull/429


abdullahwaheed:
its outdated now since this alpha is already merged in master. i think we should close this
cc: @adamstankiewicz

PR #487, opened 2023-12-14
feanil/pre scope move

feanil:
- docs: Move from restructureText to Markdown
- fix: Add a warning prior to the package move.


brian-smith-tcril (approved):

README.md
```diff
@@ -0,0 +1,303 @@
+# frontend-build
+
+## Warning: Package Moved
```
arbrandes, line 3: @abdullahwaheed alerted me to the possibility of having the warning be more "dangerous" looking, such as [here](https://www.npmjs.com/package/babel-plugin-react-intl):

image

Not sure how to do that yet, but we should probably look into it.  We can do that later, though, this is not a blocker for this PR.

README.md
```diff
@@ -0,0 +1,303 @@
...
+
+While the Paragon design system will continue to receive updates, these updates
+will no longer be published at the @edx/paragon NPM package. Instead, the
```
arbrandes, line 6:
```suggestion
will no longer be published at the @edx/frontend-build NPM package. Instead, the
```

arbrandes (approved):
Approved, save for a copy-paste mistake.

README.md
feanil: ugh, it didn't load your comment, I'll make a follow-up PR to fix this before I merge the update.

README.md
feanil: Looks like it's not too bad and I can go back and do that for the packages we've already deprecated: https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

PR #523, opened 2024-03-26
fix: update webpack/express deps

brian-smith-tcril:
a more comprehensive version of https://github.com/openedx/frontend-build/pull/522


feanil (approved):
Seems reasonable so that we don't regress back on the webpack-dev-middleware.

Issue #514, opened 2024-02-07
jest.config.js should search for both `.js` and `.jsx` extensions for `env.config`


jsnwesson:
In the likelihood that we'll need to be able to use `env.config.jsx` whenever a function is imported into the JS config, Jest will need to be able to [handle it here](https://github.com/openedx/frontend-build/blob/35415fe423a6f1103eacf1b761c003a670537726/config/jest.config.js#L7). 

For the sake of making sure bases are covered, [Webpack's common config already handles both extensions](https://github.com/openedx/frontend-build/blob/35415fe423a6f1103eacf1b761c003a670537726/config/webpack.common.config.js#L20).

PR #546, opened 2024-05-17
feat: add `ParagonWebpackPlugin` to support design tokens

dcoa:
Description:

This PR updates the original one https://github.com/openedx/frontend-build/pull/365 closer to the master branch and adds some extra tests.

Please read the original PR for additional context.

**Changes**

1.  **Delete Runtime Configuration reference:**  My reason for making this change is the nature of runtime configuration, which can override any configuration defined previously. The plugin runs during the build time and we have `env.config.js` that can set the variables.  _(let me know if my reasoning is correct)_
2. Update `@edx/paragon` to `@openedx/paragon` and add compatibility for either `@edx/brand` or `@openedx/brand` 

Finally, I would like to split this PR in 2 one for the main implementation and another one for the example app (due to this one adds frontend-platform and paragon)


openedx-webhooks:
Thanks for the pull request, @dcoa! Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation
- [Open edX discussion forum threads](https://discuss.openedx.org/)
- timeline information ("this must be merged by XX date", and why that is)
- partner information ("this is a course on edx.org")
- any other information that can help Product understand the context for the PR

All technical communication about the code itself will be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).


This is currently a draft pull request. When it is ready for our review and all tests are green, click "Ready for Review", or remove "WIP" from the title, as appropriate.

itsjeyd:
Hey @dcoa, thank you for this contribution! Please let us know when the changes are ready for review.

lib/plugins/paragon-webpack-plugin/utils.js
```diff
@@ -0,0 +1,373 @@
+const { sources } = require('webpack');
```
PKulkoRaccoonGang, line 1: [proposal]: This file is quite large. I suggest dividing it into several files and creating a folder with utilities. I think it would be a good idea to follow this structure:

**./utils**

**htmlUtils.js**
- getDescendantByTag
- findScriptInsertionPoint
- findStylesheetInsertionPoint

**scriptUtils.js**
- minifyScript
- insertScriptContentsIntoDocument

**stylesheetUtils.js**
- insertStylesheetsIntoDocument

**assetUtils.js**
- findCoreCssAsset
- findThemeVariantCssAssets
- getCssAssetsFromCompilation

**scriptContentUtils.js**
- addToScriptContents
- generateScriptContents

**urlUtils.js**
- handleVersionSubstitution
- getParagonStylesheetUrls

config/data/paragonUtils.js
```diff
@@ -0,0 +1,153 @@
+const path = require('path');
```
PKulkoRaccoonGang, line 1: [proposal]: I propose to provide for the scalability that may occur over time and divide this file into several. This will improve the code's readability and give other developers a hint to separate the code semantically.
Please let me know what your thoughts are.

lib/plugins/paragon-webpack-plugin/utils.js
```diff
@@ -0,0 +1,373 @@
...
+const parse5 = require('parse5');
+
+function getDescendantByTag(node, tag) {
```
PKulkoRaccoonGang, line 4: I also suggest adding small documentation for each function in JSDocs format

PKulkoRaccoonGang (commented):
@dcoa Thank you for re-opening and improving this PR! Left a few suggestions

lib/plugins/paragon-webpack-plugin/utils.js
dcoa: I agree and I did not the same order but let me know what you think :)

lib/plugins/paragon-webpack-plugin/utils.js
dcoa: Done :)

config/data/paragonUtils.js
dcoa: I understand your point, but I'm not entirely convinced to split this file, the structure is clear and has an order easy to follow. Also, it's focused on just one specific topic.

lib/plugins/paragon-webpack-plugin/utils.js
PKulkoRaccoonGang: Thank you! This looks great 💯

config/data/paragonUtils.js
```diff
@@ -0,0 +1,153 @@
...
+ * @typedef {Object} ParagonThemeCss
+ * @property {ParagonThemeCssAsset} core The metadata about the core Paragon theme CSS
+ * @property {Object.<string, ParagonThemeVariantCssAsset>} variants A collection of theme variants.
+ */
+
+/**
+ * Attempts to extract the Paragon theme CSS from the locally installed `@openedx/paragon` package.
+ * @param {string} dir Path to directory containing `node_modules`.
+ * @returns {ParagonThemeCss}
+ */
```
PKulkoRaccoonGang, line 44:
```suggestion
/**
 * Attempts to extract the Paragon theme CSS from the locally installed `@openedx/paragon` package.
 * @param {string} dir Path to directory containing `node_modules`.
 * @param isBrandOverride
 * @returns {ParagonThemeCss}
 */
```

config/data/paragonUtils.js
```diff
@@ -0,0 +1,154 @@
...
+ * @param {string} dir Path to directory containing `node_modules`.
+ * @returns {string} Paragon dependency version of the consuming application
+ */
+function getParagonVersion(dir, { isBrandOverride = false } = {}) {
+  const npmPackageName = isBrandOverride ? '@openedx/brand' : '@openedx/paragon';
+  const pathToPackageJson = `${dir}/node_modules/${npmPackageName}/package.json`;
+  if (!fs.existsSync(pathToPackageJson)) {
+    return undefined;
+  }
+  return JSON.parse(fs.readFileSync(pathToPackageJson)).version;
```
Alec4r, line 17: should we add the second argument to use utf8?

config/data/paragonUtils.js
```diff
@@ -0,0 +1,154 @@
...
+  if (!fs.existsSync(pathToParagonThemeOutput)) {
+    return undefined;
+  }
+  const paragonConfig = JSON.parse(fs.readFileSync(pathToParagonThemeOutput));
+  const {
+    core: themeCore,
+    variants: themeVariants,
+    defaults,
+  } = paragonConfig?.themeUrls || {};
+
```
Alec4r, line 59: should we validate if the values exist before use them?, I mean just to avoid future errors, something like if !themeCore || !themeVariants return undefined.

config/data/paragonUtils.js
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  const themeVariantResults = {};
+  validThemeVariantPaths.forEach(([themeVariant, value]) => {
+    themeVariantResults[themeVariant] = {
+      filePath: path.resolve(dir, 'node_modules', npmPackageName, 'dist', value.paths.minified),
+      entryName: isBrandOverride ? `brand.theme.variants.${themeVariant}` : `paragon.theme.variants.${themeVariant}`,
+      outputChunkName: isBrandOverride ? `brand-theme-variants-${themeVariant}` : `paragon-theme-variants-${themeVariant}`,
+    };
+  });
```
Alec4r, line 85: I think we could use .reduce instead of filtering and iterate this, it might be more efficient but I think this is an easier way to read the code.

Should we worry about efficiency?

config/data/paragonUtils.js
```diff
@@ -0,0 +1,154 @@
...
+/**
+ * @param {ParagonThemeCss} paragonThemeCss The Paragon theme CSS metadata.
+ * @returns {Object.<string, CacheGroup>} The cache groups for the Paragon theme CSS.
+ */
+function getParagonCacheGroups(paragonThemeCss) {
+  const cacheGroups = {};
+  if (!paragonThemeCss) {
+    return cacheGroups;
+  }
+  cacheGroups[paragonThemeCss.core.outputChunkName] = {
```
Alec4r, line 111: We could remove the line 107 and just define the constant from here, right?

config/data/paragonUtils.js
dcoa: This is optional, however, after reading some posts about it I think it's a good suggestion. 👍

config/data/paragonUtils.js
dcoa: It's not necessary because we are using destructuring if any of them is not defined that will return undefined by default.

dcoa:
Hi @itsjeyd yes, it is ready for review and has been tested alongside other design tokens PRs here https://github.com/openedx/frontend-app-discussions/pull/726

itsjeyd:
Thanks for confirming @dcoa.

@brian-smith-tcril @adamstankiewicz Would you be able to have a look at this PR?

brian-smith-tcril (approved):
Overall this looks like a very clean implementation, and after testing it in both https://github.com/openedx/frontend-app-discussions/pull/726 and https://github.com/openedx/frontend-app-discussions/pull/731 it seems to be working wonderfully too!

Thank you so much for this!

config/webpack.common.config.js
```diff
@@ -19,6 +46,19 @@ module.exports = {
...
+    splitChunks: {
+      chunks: 'all',
+      cacheGroups: {
+        ...getParagonCacheGroups(paragonThemeCss),
+        ...getParagonCacheGroups(brandThemeCss),
+      },
+    },
+  },
+  plugins: [
+    new RemoveEmptyScriptsPlugin(),
```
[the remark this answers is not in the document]
dcoa: Hi @brian-smith-tcril, thanks for your review, I addressed this comment and rebased the branch.

brian-smith-tcril (dismissed):

Alec4r (approved):

adamstankiewicz (commented):
I think this is pretty much ready, though I left 2 comments with questions/suggestions before I leave a 👍

PKulkoRaccoonGang (approved):

config/webpack.dev.config.js
```diff
@@ -17,12 +16,6 @@ const presets = require('../lib/presets');
 const resolvePrivateEnvConfig = require('../lib/resolvePrivateEnvConfig');
 const getLocalAliases = require('./getLocalAliases');
 
-// Add process env vars. Currently used only for setting the
-// server port and the publicPath
-dotenv.config({
-  path: path.resolve(process.cwd(), '.env.development'),
-});
-
```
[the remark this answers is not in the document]
dcoa: You are right, I intended to remove those lines in another file because they were causing some issues during the build time, and by mistake, I removed them here as well. Thanks for noticing it.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,147 @@
...
+    let envConfig;
+    if (fs.existsSync(envConfigFile)) {
+      try {
+        envConfig = require(envConfigFile);
+      } catch (error) {
+        console.info('Paragon Plugin cannot load env.config file, if you want to preload the theme url be sure your file is a CommonJS module');
+      }
+    }
+
+    const paragonThemeUrls = envConfig?.PARAGON_THEME_URLS ?? process.env.PARAGON_THEME_URLS;
```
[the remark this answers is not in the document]
dcoa: a )  It is treated as a string
b )  I think we should not encourage the use of the string of that variable because would be complicated to maintain. 

Might this one be a better approach?
```suggestion
    const paragonThemeUrls = envConfig?.PARAGON_THEME_URLS ?? {};
```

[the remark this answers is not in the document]
adamstankiewicz: I think your proposed update makes sense given `process.env.PARAGON_THEME_URLS` is parsed as a string. That is, falling back to `{}` here seems reasonable.

We can probably also remove the `resolvePrivateEnvConfig('.env.private');` towards the beginning of this file if the plugin no longer reads from `process.env` env vars 😄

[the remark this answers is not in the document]
dcoa: Y made the changes, thanks @adamstankiewicz

config/data/paragonUtils.js
```diff
@@ -0,0 +1,160 @@
...
+
+/**
+ * Attempts to extract the Paragon version from the `node_modules` of
+ * the consuming application.
+ *
+ * @param {string} dir Path to directory containing `node_modules`.
+ * @returns {string} Paragon dependency version of the consuming application
+ */
+function getParagonVersion(dir, { isBrandOverride = false } = {}) {
+  const npmPackageName = isBrandOverride ? '@openedx/brand' : '@openedx/paragon';
```
adamstankiewicz, line 12: In testing this further with the `@edx/brand` alias installed in this repo's `example` MFE app, and exposing the injected `PARAGON_THEME` global variable  within `App.jsx`, the installed brand package is not installed as this function references `@openedx/brand` rather than the aliased `@edx/brand` currently used by MFEs (e.g., [frontend-app-learning](https://github.com/openedx/frontend-app-learning/blob/3a542766d793168751a6bf48b23a4340f1ba7b2a/package.json#L33)).

I might recommend supporting the following to remain both forward-looking for the `@openedx` migration and the currently used `@edx/brand`:

* `@openedx/paragon`
* `@openedx/brand` (brand override)
* `@edx/brand` (brand override)

config/webpack.common.config.js
```diff
@@ -1,8 +1,35 @@
...
+     * {
+     *   "paragon.theme.core": "/path/to/node_modules/@openedx/paragon/dist/core.min.css",
+     *   "paragon.theme.variants.light": "/path/to/node_modules/@openedx/paragon/dist/light.min.css"
+     * }
+     */
+    ...getParagonEntryPoints(paragonThemeCss),
+    /**
+     * The entry points for the brand theme CSS. Example: ```
+     * {
+     *   "paragon.theme.core": "/path/to/node_modules/@openedx/brand/dist/core.min.css",
```
adamstankiewicz, line 28: nit: Similar to other comments; currently consuming MFEs install the brand package as `@edx/brand`.

config/webpack.common.config.js
```diff
@@ -1,8 +1,35 @@
...
+     *   "paragon.theme.variants.light": "/path/to/node_modules/@openedx/brand/dist/light.min.css"
```
adamstankiewicz, line 29: nit: Similar to other comments; currently consuming MFEs install the brand package as `@edx/brand`.

config/webpack.dev.config.js
```diff
@@ -68,43 +107,19 @@ module.exports = merge(commonConfig, {
...
-            options: {
-              postcssOptions: {
-                plugins: [
-                  PostCssAutoprefixerPlugin(),
-                  PostCssRTLCSS(),
-                  PostCssCustomMediaCSS(),
-                ],
-              },
-            },
+            resource: /(@openedx\/paragon|@edx\/brand)/,
```
adamstankiewicz, line 112: nit: if other code paths are supporting `@openedx/brand` as the NPM alias (which afaik MFEs do not yet use), does this regex support `@openedx/brand`, too? It currently only supports `@edx/brand` while other code paths only supports `@openedx/brand`.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,141 @@
...
+} = require('../../../config/data/paragonUtils');
+const {
+  injectMetadataIntoDocument,
+  getParagonStylesheetUrls,
+  injectParagonCoreStylesheets,
+  injectParagonThemeVariantStylesheets,
+} = require('./utils');
+
+// Resolve configuration `env.config`.
+const envConfigFile = path.resolve(process.cwd(), 'env.config.js');
```
adamstankiewicz, line 16: [nit/inform] likely OK for now, but we might want to follow-up here to ensure we support `env.config` files defined with the `.jsx` extension (related to https://github.com/openedx/frontend-build/issues/514).

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,141 @@
...
+      this.injectParagonStylesheetsIntoDocument,
+      ...processAssetsHandlers,
+    ].map(handler => handler.bind(this));
+  }
+
+  /**
+   * Resolves the MFE configuration for ``PARAGON_THEME_URLS`` in the following priority order:
+   *
+   * 1. ``envConfig.PARAGON_THEME_URLS``
+   * 2. ``process.env.PARAGON_THEME_URLS``
```
adamstankiewicz, line 47: nit: We are no longer attempting to support environment variable based approach to defining `PARAGON_THEME_URLS` due to the string parsing issue.

adamstankiewicz (changes requested):
In addition to a few more minor nits/suggestions/deferred follow-up opportunities, I noticed a bug that when attempting to verify the behavior with the `@edx/brand` NPM alias currently used by MFEs, the plugin does not recognize the brand package because it only looks for `@openedx/brand`, which MFEs / consumers have not migrated to using as the standard alias yet. 


In the below screenshot, note there are no preloaded `link` elements for the configured/installed `@edx/brand` package's `core.min.css` or `light.min.css`:

image

tl;dr; Requesting changes since the brand related behavior currently does not seem to be working as expected when using the NPM alias `@edx/brand`.

config/data/paragonUtils.js
```diff
@@ -0,0 +1,160 @@
...
+ */
+
+/**
+ * Attempts to extract the Paragon theme CSS from the locally installed `@openedx/paragon` package.
+ * @param {string} dir Path to directory containing `node_modules`.
+ * @param {boolean} isBrandOverride
+ * @returns {ParagonThemeCss}
+ */
+function getParagonThemeCss(dir, { isBrandOverride = false } = {}) {
+  const npmPackageName = isBrandOverride ? '@openedx/brand' : '@openedx/paragon';
```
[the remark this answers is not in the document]
dcoa: The reason why `@edx/brand` is not considered is that we would like to encourage the migration of that alias as much as possible ([additional context](https://openedx.slack.com/archives/C071U9E4VNV/p1715940913033459)), that is why is included in the list of changes in the PR description,  as you can see in the [example app PR](https://github.com/openedx/frontend-build/pull/549/files#diff-a593fc14fdef41aaf092f774b92cb74bafeda3e9559f8ae4762374bf4af80c92R18) 

Even the current readme for the repository makes reference to `@openedx/brand` instead of `@edx/brand`.

https://github.com/openedx/frontend-build/blob/ae8ce98d5252d9ddd4e730d6b21ca40257453bef/README.md?plain=1#L158

I consider that we could be more explicit about this change in the [MFE Migration Guide](https://openedx.atlassian.net/wiki/spaces/BPL/pages/3770744958/Migrating+MFEs+to+Paragon+design+tokens+and+CSS+variables) instead. What do you think?

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
dcoa: I agree with this and I tried to use `.jsx` files,  and found the following scenario:

Due to Node only supporting js, json natively,  read jsx should be done with fs.readFile and processing the result (string) to find the `PARAGON_THEME_URLS`. 

The reason why I don't use it is that I don't want to increase the build time more processing the file.

adamstankiewicz: There are MFEs that currently assume the `.jsx` extension works and uses it ([source](https://github.com/search?q=org%3Aopenedx%20env.config.jsx&type=code)), as well as private `env.config.jsx` files currently used to introduce private JSX-based plugins/widgets (e.g., for at least 2U's instance of `frontend-app-learner-record`).

Given this, I'd be concerned that without `ParagonWebpackPlugin` supporting `.jsx`, any MFEs that define custom JSX-based plugins (e.g., via `@openedx/frontend-plugin-framework`) with `env.config.jsx` won't be able to configure `PARAGON_THEME_URLS`.

FWIW, an alternative to parsing the `env.config.jsx` as a string with `fs.readFile` would be to conditionally rely on `@babel/register` to use Babel to transform JSX to vanilla JS during the `require` import.

**Example solution with `@babel/register`**

1. Install `@babel/register` in root `package.json`.

2. Define a file `register-babel.js` with the following contents (e.g., under `lib` directory):

```js
require('@babel/register')({
  presets: ['@babel/preset-env', '@babel/preset-react'],
});
```

3. Update `ParagonWebpackPlugin` to check against supported extensions (`js` and `jsx`).

```js
const supportedEnvConfigExtensions = ['js', 'jsx'];
const envConfigFile = path.resolve(process.cwd(), 'env.config');
```

```js
async resolveParagonThemeUrlsFromConfig() {
  if (!supportedEnvConfigExtensions.some(ext => fs.existsSync(`${envConfigFile}.${ext}`))) {
    // env.config file does not exist
    return;
  }

  let envConfig;
  require('../../register-babel');
  try {
    envConfig = require(envConfigFile);
  } catch (error) {
    console.info('ParagonWebpackPlugin cannot load env.config file. Skipping...');
  }

  // ...
}
```

adamstankiewicz: > The reason why I don't use it is that I don't want to increase the build time more processing the file.

If existing MFEs / frontend-platform is compatible with JSX-based `env.config` files, I feel all usages of `env.config` should support the same extensions, even if it means slightly more processing at build time.

Otherwise, it's fairly confusing for consumers/engineers to know which `env.config` file extensions work for different parts of the system (e.g., can use JSX-based `env.config` for private plugins but not for `PARAGON_THEME_URLS`).

brian-smith-tcril:
@dcoa in https://openedx.slack.com/archives/C071U9E4VNV/p1715966767660459?thread_ts=1715940913.033459&cid=C071U9E4VNV I mentioned
> I'd prefer to move over to using `@openedx/brand-openedx` everywhere but I don't remember all of the gotchas/blockers around that. I think moving forward we should try to use `@openedx` but if we run into conflicts because of it we can alias back to `@edx` and note why that was required in any given instance.

When I wrote that I wasn't thinking about landing this PR as a non breaking change. Since the plan now is to land this as a non breaking change I think that fully justifies keeping the `@edx` namespace for now.

I would still like to make the move to the `@openedx` namespace, but since doing so is a breaking change I feel it would be better off as a separate PR so as to not delay the merge of this one.

adamstankiewicz:
Thanks @brian-smith-tcril. Only thing I'd add regarding the brand package scope (`@edx/brand` vs. `@openedx/brand`) is that I think it might make sense to handle _either_ scope so we are compatible with the current state of MFEs' usage with `@edx/brand`, but also be forward-looking knowing we will migrate to `@openedx/brand` eventually. Once we're confident consumers/instances have fully migrated to `@openedx/brand`, we could come back to remove the `@edx/brand` at that point. Supporting both scopes now would eliminate the need for any coordinated effort in adding support for `@openedx/brand` down the road.

config/data/paragonUtils.js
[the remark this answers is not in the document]
adamstankiewicz: > Even the current readme for the repository makes reference to @openedx/brand instead of @edx/brand.

To expand on @brian-smith-tcril's response below regarding keeping parity with the existing `@edx/brand` scope used today given the non-breaking nature of this release, I believe the README prematurely referenced `@openedx/brand` as part of [this PR](https://github.com/openedx/frontend-build/pull/490) given that migration of NPM scope for the brand packages has not yet been prioritized/executed. Looks like the brand package was incorrectly grouped with the scope migrations for other NPM packages (i.e., `@openedx/paragon` and `@openedx/frontend-build`).

By having the current README recommend `@openedx/brand` in the `module.config.js` within the README would never be resolved as the alias since `@openedx/brand` is not used in consuming apps, so it's currently incorrect documentation.

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
dcoa: Just to be clear `PARAGON_THEME_URLS` can still be implemented in  JSX-based configuration because `frontend-platform`  can read it so consumers/engineers can use it.

Thanks for the explanation about `register-babel` I will take a look.

adamstankiewicz: > Just to be clear PARAGON_THEME_URLS can still be implemented in JSX-based configuration because frontend-platform can read it so consumers/engineers can use it.

While true, not supporting JSX-based `env.config` in frontend-build comes with a performance tradeoff for users. If frontend-build does not support JSX-based `env.config` for defining `PARAGON_THEME_URLS`, the associated `<link rel="preload">` elements added to the HTML document by frontend-build would be missing from the HTML document.

Without the preloaded stylesheets ([reference about preload](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload)), users will need to wait to download/parse/execute the MFE's JS (including `@edx/frontend-platform`'s source) _before_ the `<link>` elements would be created / download started for the configured `PARAGON_THEME_URLS` defined by `env.config.jsx`.

Given `AppProvider` in frontend-platform blocks the rendering of the MFE's contents until the Paragon CSS is loaded ([source](https://github.com/openedx/frontend-platform/pull/689/files#diff-e986d374444ced0f5162cc6cb8b3dac6cbd228f339851f2a99a4aec545559d87R89-R91)), I feel it's important that these `<link>` elements are preloaded on initial page load regardless of using `env.config.js` or `env.config.jsx`, without being blocked by downloading JavaScript assets first to ensure this solution is as performant as possible for users.

config/webpack.dev.config.js
dcoa: This is solve @adamstankiewicz @brian-smith-tcril

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
dcoa: I have been testing `@babel/register` and works when the file is `jsx` but not alongside `@openedx/frontend-plugin-framework` (fail with `SyntaxError: Unexpected token 'export'` trying to resolve the plugin dependencies). 

I am testing if some plugin can help in that case

dcoa: I couldn't find a solution for the dependencies issue, @adamstankiewicz do you have an idea that could work? I'm using the file in learning MFE for testing leaving only `@openedx/frontend-plugin-framework` the example for footer-slot and PARAGON_THEME_URLS

```jsx
import { PLUGIN_OPERATIONS, DIRECT_PLUGIN } from '@openedx/frontend-plugin-framework';

// Load environment variables from .env file
const config = {
  ...process.env,
  pluginSlots: {
    footer_slot: {
      plugins: [
        {
          // Hide the default footer
          op: PLUGIN_OPERATIONS.Hide,
          widgetId: 'default_contents',
        },
        {
          // Insert a custom footer
          op: PLUGIN_OPERATIONS.Insert,
          widget: {
            id: 'custom_footer',
            type: DIRECT_PLUGIN,
            RenderWidget: () => (
              <h1 style={{ textAlign: 'center' }}>🦶</h1>
            ),
          },
        },
      ]
    }
  },
  PARAGON_THEME_URLS: {
    core: {
      urls: {
        default: 'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/core.min.css',
      },
    },
    defaults: {
      light: 'light',
    },
    variants: {
      light: {
        urls: {
          default: 'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/light.min.css',
        },
      },
    },
  },
};

export default config;
```

The import error persists.

brian-smith-tcril: @dcoa do you have a branch or a diff somewhere with what you've tried in regards to `@babel/register`? I find it's often easier to talk about code when there's a diff to look at.

Also, I did a bit of searching and I wasn't able to find much about the error you're seeing outside of https://stackoverflow.com/a/54663822. Someone resolved a similar issue by setting a root directory for babel, but I'm not sure if that applies at all here.

dcoa: [Just for the record] [This one](https://gist.github.com/dcoa/f53caa609fd9f817c87baee038e4a528) compiles the different config that I use to test `@babel/register`

dcoa: @adamstankiewicz @brian-smith-tcril  as mentioned in Slack, I added the proposed workaround to support `process.env.PARAGON_THEME_URLS`. 

We can storage the variable in JSON format but should be a minified version of it (1 line)

```env
PARAGON_THEME_URLS= {core:{urls:{default:'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/core.min.css'}},defaults:{light:'light'},variants:{light:{urls:{default:'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/light.min.css'}}}}

```

lib/plugins/paragon-webpack-plugin/ParagonWebpackPlugin.js
```diff
@@ -0,0 +1,139 @@
...
+      this.injectParagonStylesheetsIntoDocument,
+      ...processAssetsHandlers,
+    ].map(handler => handler.bind(this));
+  }
+
+  /**
+   * Resolves the MFE configuration for ``PARAGON_THEME_URLS`` in the following priority order:
+   *
+   * 1. ``envConfig.PARAGON_THEME_URLS`` (JS based)
+   * 3. ``process.env.PARAGON_THEME_URLS`` (JSON based)
```
brian-smith-tcril, line 47: I feel like having `PARAGON_THEME_URLS` work in an `env.config.js` file but not an `env.config.jsx` file is more confusing than just saying `PARAGON_THEME_URLS` *must* be set as an environment variable.

I'd love to hear what others think about this!

davidjoy: To be clear - this file _cannot_ import or use `env.config` with any file extension, and we must use `process.env` to get this value.  The env.config file is _runtime_ configuration and is expected to import runtime files that will require transpilation (i.e., jsx/tsx) and depend on their own runtime dependencies.  That code can't meaningfully pollute the build-time configuration, and having multiple files named env.config with different extensions would be seriously confusing for folks.  

At a later date we may decide to create a new type of build-only configuration file, but we're just not there at the moment.  For now, we just need to use process.env either via command line environment variables or the `.env` files, and it'll come in as a string we need to `JSON.parse`.

davidjoy: Furthermore, worse, would be for build-time code and build-time dependencies to find their way into env.config... that would mean build-time code gets bundled with the app for a browser environment which is a huge no-no and would definitely break.

adamstankiewicz: +1. Agreed, we should only be relying on `process.env.PARAGON_THEME_URLS` at this point, with `JSON.parse` for the build-time vs. runtime configuration that @davidjoy is alluding to above (across both this frontend-build PR _and_ within frontend-platform).

davidjoy: On the runtime side for frontend-platform's use, by all means copy PARAGON_THEME_URLS into env.config and format it nicely as an actual object instead of a string.  But for now it does need to be duplicated, unfortunately, because the two types of config can't mingle.

arbrandes: I gave a couple of thumbs up above, but better to be explicit: I agree that for now, `process.env.PARAGON_THEME_URLS` should be what is used, and only that.

I also like the idea of eventually having a separate build-only config file, thus doing away with any potential for confusion or breakage.  (While at the same time finally removing support for the .env files, hopefully.)

brian-smith-tcril: > I also like the idea of eventually having a separate build-only config file, thus doing away with any potential for confusion or breakage. (While at the same time finally removing support for the .env files, hopefully.)

### :100: 

That lines up perfectly with the feelings expressed in [the slack thread](https://openedx.slack.com/archives/C04BM6YC7A6/p1723069071930839?thread_ts=1723049201.030959&cid=C04BM6YC7A6).

dcoa: Done 😁

adamstankiewicz (approved):
LGTM :shipit:  Great job! I also closed the original POC implementation's [PR](https://github.com/openedx/frontend-platform/pull/440).

openedx-webhooks:
@dcoa 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/frontend-build&entry.867055334=https://github.com/openedx/frontend-build/pull/546&entry.1484655318=https://github.com/dcoa&entry.752974735=2024-05-17+05:08&entry.1917517419=2024-08-13+21:21&entry.2133058324=Yes) so we can improve your experience in the future.

davidjoy:
Hmm.  @adamstankiewicz @dcoa - 

I'm pulling this work into openedx/frontend-base and just noticed the key `default` in this data structure:

```
PARAGON_THEME_URLS: {
    core: {
      urls: {
        default: 'https://cdn.jsdelivr.net/npm/@openedx/paragon@alpha/dist/core.min.css',
      },
    },
  // ...
}
```

Don't know about anyone else, but Typescript and my IDE are freaking out that we used a keyword as a object field name. 😬  

I seriously wonder if we want to amend that to use a different word instead of 'default'...

davidjoy:
I had a comment here about using `default` as a key name, but I think it's actually fine.

PR #565, opened 2024-06-20
fix: bump shared eslint config version

adamstankiewicz:
The shared `@edx/eslint-config` was recently updated to support the `es2020` environment such that syntax like `globalThis` won't cause linting errors.

This PR bumps the version of `@edx/eslint-config` to pull in this change for consuming MFEs and shared JS libraries (e.g., frontend-platform).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@babel/plugin-syntax-dynamic-import": "7.8.3",
         "@babel/preset-env": "7.24.7",
         "@babel/preset-react": "7.24.7",
-        "@edx/eslint-config": "4.0.0",
+        "@edx/eslint-config": "4.1.0",
         "@edx/new-relic-source-map-webpack-plugin": "2.1.0",
         "@edx/typescript-config": "1.0.1",
         "@formatjs/cli": "^6.0.3",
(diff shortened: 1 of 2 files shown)


bradenmacdonald (approved):
Looks great! I tested this in frontend-platform and frontend-app-course-authoring. After this bump, it stopped showing `globalThis` as an eslint error.

bradenmacdonald:
BTW, something I mentioned to @davidjoy recently: I don't understand why we even have `eslint-config` as a separate repo. It contains 16 boilerplate files and 1 useful `[.eslintrc.js](https://github.com/openedx/eslint-config/blob/master/.eslintrc.js)` file which could just as easily be merged into this repo's `config/.eslintrc.js` file. Plus, `eslint-config` and `frontend-build` specify different versions of eslint and various dependencies.

adamstankiewicz:
> BTW, something I mentioned to @davidjoy recently: I don't understand why we even have `eslint-config` as a separate repo. It contains 16 boilerplate files and 1 useful [`.eslintrc.js`](https://github.com/openedx/eslint-config/blob/master/.eslintrc.js) file which could just as easily be merged into this repo's `config/.eslintrc.js` file.
> 
> And according to npm, [it's only used by `frontend-build`](https://www.npmjs.com/package/@edx/eslint-config?activeTab=dependents) anyways.

`edx/eslint-config` predates `@openedx/frontend-build` by awhile. `@edx/eslint-config` is still consumed by other frontend-facing repos (e.g., [edx-platform](https://github.com/openedx/edx-platform/blob/ad368de72a8edc4c8e4f26889258018ce5ea9a06/.eslintrc.json#L2), [studio-frontend](https://github.com/openedx/studio-frontend/blob/357e7e635ce1e3260e2e7a5abcc5e2b0c786cf4f/.eslintrc.json#L2), [credentials](https://github.com/openedx/credentials/blob/557a0e42cc17d77e5298c8a311f4b97ebec16114/.eslintrc.json#L2), [Paragon](https://github.com/openedx/paragon/blob/83bad6064506d7b6456cda235177e8d2eaf50333/.eslintrc.js#L5)), which don't rely on `@edx/frontend-build`.

The ESLint config in `@openedx/frontend-build` specifically is intended only for MFEs and more recent shared JS libraries that consume it. That said, it does seem like there's a few non-legacy repos still referring to `@edx/eslint-config` when they should be relying solely the base config from `@edx/frontend-build` ([example](https://github.com/openedx/frontend-app-learner-portal-programs/blob/70215dce61789f869211091662e1c15f65391a5f/.eslintrc#L2)).

> Plus, eslint-config and frontend-build specify different versions of eslint and various dependencies.

According to the `peerDependencies` in `@edx/eslint-config`, it supports both v7 and v8 of ESLint; `@openedx/frontend-build` relies on v8 of ESLint. As long as the `peerDependencies` can be fulfilled, I'm not sure there's much concern around slightly different ESLint versions between these two packages, so long as they are compatible with each other.

brobro10000 (approved):

bradenmacdonald:
I see now. Thanks for the context on that, @adamstankiewicz !

PR #609, opened 2024-11-08
chore: bump eslint

arbrandes:
Bump eslint so that we get the new rules that were merged in https://github.com/openedx/eslint-config/pull/164.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@babel/plugin-syntax-dynamic-import": "7.8.3",
         "@babel/preset-env": "7.24.8",
         "@babel/preset-react": "7.24.7",
-        "@edx/eslint-config": "4.2.0",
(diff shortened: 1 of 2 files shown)


bradenmacdonald (approved):
Thanks! Do you know why renovate didn't open this PR automatically? Maybe because it was "4.2.0" instead of "^4.2.0" ?

arbrandes:
@bradenmacdonald, yeah, we were missing the caret.  I don't know for sure, but I figure Renovate would respect that.

Anyway, it's conspicuously missing from most other deps as well.  I didn't do any prospecting to find out why that is, but I figure at least our internal deps should have it.

davidjoy:
Last I knew, Renovate intentionally removes the ^ in an initial PR so that it can fully manage the versions. So... by design, at least as originally implemented in our repos.

PR #636, opened 2025-03-18
feat: ensure Webpack and ESLint adhere to `paths` defined in `tsconfig.json`

adamstankiewicz:
Some MFEs have opted to simplify their imports by setting up aliases to avoid relative paths.

Example:

```js
import MyComponent from '../path/to/MyComponent';
```

versus

```js
import MyComponent from '@components/MyComponent';
```

Currently, MFEs need to do the following 3 things to enable this:

1. Define `paths` in `tsconfig.json`.
2. Extend the base Webpack config from `@openedx/frontend-build` to define the `alias` option, corresponding to the `paths` from (1).
3. Extend the base ESLint config to ensure `import/resolver` acknowledges `tsconfig.json` OR the configured Webpack `alias` option.

This current approach to alias imports is brittle in that it relies on manual verification that the `paths` in `tsconfig.json` match the Webpack `alias` options. If these paths differ (e.g., a new path is only added to `tsconfig.json`), the import will not recognized by Webpack properly and fail the build.

This PR proposes using `tsconfig.json` (if available) as the source-of-truth for which aliases exist across the build tooling (i.e., TypeScript, Webpack, ESLint) vs. defining them in multiple places.

To do this, the base Webpack and ESLint configs are extended to integrate with:
* `tsconfig-paths-webpack-plugin`. Ensures Webpack adheres to the configured `paths` in `tsconfig.json`.
* `eslint-import-resolver-typescript`. Ensures ESLint can properly resolve the aliases to the configured `paths` in `tsconfig.json`.

With these changes, consuming MFEs opting to define custom `paths` in their `tsconfig.json` files will only need to define the `paths`, and the Webpack/ESlint tooling should work out-of-the-box.


.eslintrc.js
```diff
@@ -7,4 +7,5 @@ module.exports = createConfig('eslint', {
     'global-require': 'off',
     'no-template-curly-in-string': 'off',
   },
+  ignorePatterns: ['example/**/*'],
```
adamstankiewicz, line 10: [inform] This root `.eslintrc.js` is no longer responsible for linting the `example` MFE app containing in this project. Instead, the `example` MFE app should be responsible for its linting config, given it has its own `.eslintrc.js` config file.

The root `lint` NPM script is modified below to ensure CI runs the `lint` command for both the root and the example MFE app.

dcoa:
Hi, @adamstankiewicz  I was updating my branch of frontend-app-learner-dashboard and the build failed because webpack cannot find the `tsconfig.json` file as you can see in the [logs](https://github.com/openedx/frontend-app-learner-dashboard/actions/runs/14122223957/job/39564410662?pr=569)

adamstankiewicz:
Looking into it, thanks @dcoa.

adamstankiewicz:
@dcoa The fix #643 has been merged/released to v14.4.1. Please let met know if you continue to have issues with the upgrade. Apologies for the inconvenience!

PR #646, opened 2025-04-01
feat: get rid of bundle analyzer on by default in prod

regisb:
Every run of `npm run build` generates a report.html file that includes file size stats. It's useless in most cases. We suggest to remove it (along with the dependency) in all MFEs.

This shaves 14s (out of 174s) of the build on my laptop for the learning MFE build.

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -2,7 +2,6 @@
 // optimized bundles at the expense of a longer build time.
 
 const ImageMinimizerPlugin = require('image-minimizer-webpack-plugin');
-const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 const { merge } = require('webpack-merge');
 const CssNano = require('cssnano');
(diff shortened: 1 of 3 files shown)


regisb:
This is ready for review.

adamstankiewicz:
> It's useless in most cases.

I personally refer to the generated `report.html` about production builds often in the MFEs I support to keep an eye on the application's bundle size, find opportunities for code splitting, see which `node_modules` packages are largest, etc., so I personally wouldn't necessarily say its "useless in most cases." Removing this for all MFEs does not help to encourage the good practice of keeping an eye on and improving bundle size.

That said, I could see the generation of the report perhaps being more an opt-in option when creating a build, e.g. `fedx-scripts webpack --create-bundle-report` or similar so it doesn't impact CI builds but could still be selectively used during local development.

regisb:
The generation of report.html in production builds has the unfortunate side effect that it's also available on edx.org... See for instance: https://authn.edx.org/report.html

itsjeyd:
@regisb Thanks for sharing your thoughts :D I'll close this PR now.

PR #650, opened 2025-04-10
feat(formatjs): support additional src folders

igobranco:
Enhance the `fedx-scripts.js` script on `formatjs` with an option to extract translations on additional source folders.

Example:
`--include=plugins --include=plugins2`

This PR is required to fix the extraction process on the https://github.com/openedx/frontend-app-authoring/ because there are an additional [`plugins` source folder](https://github.com/openedx/frontend-app-authoring/tree/master/plugins/course-apps).

Then the next step is to change the `i18n_extract` on package.json to include the additional folder on https://github.com/openedx/frontend-app-authoring/blob/04faf54ad8ca74a288c872f1140eafc0aa4f9763/package.json#L14

From:
```json
"i18n_extract": "fedx-scripts formatjs extract",
```

To:
```json
"i18n_extract": "fedx-scripts formatjs extract --include=plugins",
```

Additionally it was added a new commit to extract translations from `.ts` files.

## Additional information
https://github.com/openedx/frontend-app-authoring/pull/1589

### DIFF
--- a/bin/fedx-scripts.js
+++ b/bin/fedx-scripts.js
@@ -69,8 +69,12 @@ switch (commandName) {
     require('webpack-dev-server/bin/webpack-dev-server');
     break;
   case 'formatjs': {
-    // To extract more messages on other source folders use:
-    // --include=plugins --include=plugins2
+    // The include option is used to specify which additional source folders to extract messages from.
+    // To extract more messages on other source folders use: --include=plugins --include=plugins2
+    // The intention use case is to allow extraction from the 'plugins' directory on 'frontend-app-authoring'.
+    // That plugins folder were kept outside the src folder to ensure they remain independent and
+    // can function without relying on the MFE environment's special features.
+    // This approach allows them to be packaged separately as NPM packages.
     const additionalSrcFolders = [];
     process.argv.forEach((val, index) => {
       // if val starts with --include= then add the value to additionalSrcFolders
--- a/bin/fedx-scripts.js
+++ b/bin/fedx-scripts.js
@@ -88,8 +88,9 @@ switch (commandName) {
     const commonArgs = [
       '--format', 'node_modules/@openedx/frontend-build/lib/formatter.js',
       '--ignore', `${srcFoldersString}/**/*.json`,
+      '--ignore', `${srcFoldersString}/**/*.d.ts`,
       '--out-file', './temp/babel-plugin-formatjs/Default.messages.json',
-      '--', `${srcFoldersString}/**/*.js*`,
+      '--', `${srcFoldersString}/**/*.{j,t}s*`,
     ];
     process.argv = process.argv.concat(commonArgs);
     ensureConfigOption(presets.formatjs);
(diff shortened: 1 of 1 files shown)


igobranco:
FYI @OmarIthawi

igobranco:
> This approach seems like a reasonable way to add additional directories to extract, but I am curious about the motivation behind this. I'm mostly wondering why the `plugins` directory in `frontend-app-authoring` isn't in `src`.

@brian-smith-tcril
I'm not an expert on the `frontend-app-authoring`...
Do you think that I should include your proposal and my ignore the `*.d.ts` files on this PR?
https://github.com/openedx/frontend-app-authoring/pull/1801#issuecomment-2792512538

brian-smith-tcril:
> Do you think that I should include your proposal and my ignore the `*.d.ts` files on this PR? [openedx/frontend-app-authoring#1801 (comment)](https://github.com/openedx/frontend-app-authoring/pull/1801#issuecomment-2792512538)

That sounds good to me! Thanks!

OmarIthawi:
Thanks @igobranco and sorry for keeping you jumping between PRs. This is the nature of opensource contribution in opensource projects, but I with a bit of effort we'll get this improvement for everyone and it'll be a happy moment when we get your PR merged.

As a long time contributor, I face this issue almost for all non-trivial changes. So I appreciate you pushing through this contribution.

Thanks for @brian-smith-tcril for the review :)

igobranco:
@OmarIthawi @brian-smith-tcril 

I've added a new commit to this PR, with the @brian-smith-tcril suggestion.
I had to skip the `*.d.ts` files.

The increase of extracted translations is big on both authoring and learning MFEs, but on learner-dashboard it don't extract any more more strings.

Running the command before and after the change:

```bash
npm run i18n_extract && grep '"id"' temp/babel-plugin-formatjs/Default.messages.json | wc -l
```

| project | from | to |
| -- | -- | -- |
| frontend-app-authoring |  1730  | 2309 |
| frontend-app-learning |  126 | 499 |
| frontend-app-learner-dashboard | 144 | 144 |

I tested on 3 repositories and it works, at least don't crash!

bradenmacdonald:
> I am curious about the motivation behind this. I'm mostly wondering why the `plugins` directory in `frontend-app-authoring` isn't in `src`.

That was my decision, and I don't remember if it was necessary for technical reasons or not. But I think the main motivation was this: if the plugins are in the `src` folder, they may benefit from special features of the MFE environment and build system, and when someone comes along and tries to build a third-party plugin that's not part of the MFE, it may not be able to achieve similar things because it doesn't live in the `src` directory. So keeping them outside of `src` was a way of ensuring that their code was relatively independent, at least in theory. The entire course-apps plugins system of the `authoring` MFE is very idiosyncratic and experimental though, so at some point I'd like to find a better approach.

brian-smith-tcril:
Thanks for providing that context @bradenmacdonald! With that in mind, are you happy with the syntax this PR adds? I think `--include=myAdditionalDirectory` seems reasonable, but I'd love to hear your thoughts on the matter!

brian-smith-tcril:
A comment about the use case sounds great!

@igobranco would you mind adding a comment mentioning the `plugins` directory in `frontend-app-authoring` specifically?

igobranco:
Added a new commit with that comment.

brian-smith-tcril (approved):
:+1:

bradenmacdonald (approved):
Thanks a lot for figuring this out!

